PR #16, opened 2026-06-20
Automatic Cloudflare quick tunnel for `devspace serve` + styled startup UI

prestonlogan:
## What this adds

Today `devspace init` requires the user to stand up their own tunnel/reverse proxy and paste the public URL by hand. This PR adds an **opt-in "Automatic Cloudflare quick tunnel"** mode so `devspace serve` can expose itself publicly on its own — no second terminal, no manual URL — plus a [clack](https://github.com/bombshell-dev/clack)-styled startup summary that matches the existing setup screens.

It's fully opt-in and the manual-URL flow is unchanged.

## Behavior

**`devspace init`** now asks how the host should be reached:
- **Automatic Cloudflare quick tunnel (recommended)** → persists `"tunnel": "cloudflare"` in `config.json`
- **Manual public URL** → existing behavior, unchanged

**`devspace serve`** when the tunnel is enabled:
1. Resolves `cloudflared` from `PATH`, then `~/.devspace/bin`, auto-installing the official release if missing.
2. Runs `cloudflared tunnel --url http://<host>:<port> --no-autoupdate` and captures the `https://<random>.trycloudflare.com` URL.
3. Sets `DEVSPACE_PUBLIC_BASE_URL` so the tunnel hostname is added to the Host-header allowlist via the existing `loadConfig()` path.
4. Prints a styled summary box and tears the tunnel down on `SIGINT`/`SIGTERM`/`exit`.

**Per-run overrides:** `devspace serve --tunnel` / `--no-tunnel`, or `DEVSPACE_TUNNEL=cloudflare|none`. `CLOUDFLARED_BIN` points at a specific binary.

## Example

```
┌  DevSpace
◇  Cloudflare tunnel ready  https://poultry-chevy-console-criteria.trycloudflare.com
◇  Server running (Cloudflare tunnel live) ──────────────────────────────╮
│  Local   http://127.0.0.1:7676/mcp                                      │
│  Public  https://poultry-chevy-console-criteria.trycloudflare.com/mcp   │
│  Roots   /Users/me/Development                                          │
│  Auth    Owner password approval required                               │
╰─────────────────────────────────────────────────────────────────────────╯
└  Press Ctrl+C to stop. Keep this terminal open — open a new tab for other work.
```

## Files
- **`src/cloudflare-tunnel.ts`** (new) — cloudflared resolve/install + quick-tunnel lifecycle. All spawns use `shell: false`.
- **`src/cli.ts`** — init tunnel-mode prompt, tunnel-aware `serve()`, styled output, `--tunnel`/`--no-tunnel`, help text.
- **`src/user-config.ts`** — `DevspaceUserConfig.tunnel?: "cloudflare"`.

## Security considerations
- All `spawn`/`spawnSync` calls use `shell: false` with array args; the only interpolated value (`localBaseUrl`) is built from validated config and passed as a single argv element — no shell/arg injection.
- The release asset is chosen from a fixed platform/arch allowlist (never user input); install dir is `0o700`.
- The captured public URL is matched by a tight regex (`https://[a-z0-9-]+\.trycloudflare\.com`), so only a real trycloudflare hostname can be set, and only that exact host is added to the allowlist (no wildcard). OAuth owner-password approval is unchanged.
- **Known tradeoff:** the downloaded `cloudflared` is fetched over HTTPS from the official `cloudflare/cloudflared` releases and smoke-tested with `--version`, but is **not checksum-pinned** (trust-on-first-use — the same trust model as running `cloudflared` from `PATH`). Glad to add SHA-256 verification if you'd prefer.

## Testing
- `npm run typecheck`, `npm run build`, `npm test` all pass.
- Exercised end-to-end on macOS arm64: install/resolve, quick-tunnel URL capture, Host-allowlist wiring, and teardown on Ctrl+C.

Happy to adjust naming, defaults (e.g. keep it strictly flag-gated rather than persisted), or split the styled-UI change into its own PR if you'd rather keep this focused on the tunnel.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -14,10 +14,23 @@ import {
   type DevspaceUserConfig,
 } from "./user-config.js";
 import { expandHomePath } from "./roots.js";
+import { startQuickTunnel, type QuickTunnel } from "./cloudflare-tunnel.js";
 
 type Command = "serve" | "init" | "doctor" | "config" | "help";
 const require = createRequire(import.meta.url);
 const SUPPORTED_NODE_RANGE = ">=20.12 <27";
+
+const useColor = Boolean(output.isTTY) && !process.env.NO_COLOR;
+const paint = (code: string) => (value: string) =>
+  useColor ? `\x1b[${code}m${value}\x1b[0m` : String(value);
+const c = {
+  bold: paint("1"),
+  dim: paint("2"),
+  cyan: paint("36"),
+  green: paint("32"),
+  yellow: paint("33"),
+  magenta: paint("35"),
+};
 
 async function main(argv: string[]): Promise<void> {
   assertSupportedNode();
@@ -28,7 +41,7 @@ async function main(argv: string[]): Promise<void> {
   switch (command) {
     case "serve":
       await ensureConfigured();
-      await serve();
+      await serve(args);
       return;
     case "init":
       await runInit({ force: args.includes("--force") });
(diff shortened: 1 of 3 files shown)


Waishnav:
This is really nice idea which i also thought about, but i don't want users to force their tunneling service to cloudflare tunnel. Unless cloudflare sponsors us.

Lets hope, it happens in upcoming days.

PR #19, opened 2026-06-20
Persist OAuth state across restarts

welearner:
## Summary

This PR persists OAuth clients and refresh tokens to disk so OAuth sessions can survive DevSpace server restarts.

Access tokens and authorization codes remain in memory only.

## Motivation

DevSpace currently stores OAuth state in memory. When the server restarts, registered OAuth clients and refresh tokens are lost, requiring clients to repeat the OAuth registration and authorization flow.

Persisting clients and refresh tokens allows clients to recover after a restart using the standard OAuth refresh flow.

## Changes

- Persist registered OAuth clients.
- Persist refresh tokens as token hashes.
- Restore non-expired refresh tokens on startup.
- Keep access tokens in memory only.
- Keep authorization codes in memory only.
- Add `DEVSPACE_OAUTH_STATE_PATH` configuration.
- Default OAuth state path to `<DEVSPACE_STATE_DIR>/oauth.json`.
- Add tests for OAuth state persistence and refresh-token recovery after restart.

## Security notes

- Raw token values are not written to disk.
- Refresh tokens are stored as hashes.
- Expired refresh tokens are ignored when loading state.
- Access tokens are intentionally not persisted because they are short-lived bearer credentials.

## Testing

- `npm test`
- `npm run typecheck`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -26,6 +26,8 @@ DEVSPACE_LOG_ASSETS=0
 DEVSPACE_LOG_TOOL_CALLS=1
 DEVSPACE_LOG_SHELL_COMMANDS=0
 # DEVSPACE_TRUST_PROXY=1
+# OAuth client registrations and refresh tokens are stored in
+# $DEVSPACE_STATE_DIR/oauth.json. Access tokens remain in memory only.
 # DEVSPACE_STATE_DIR=/home/waishnav/.local/share/devspace
 # DEVSPACE_WORKTREE_ROOT=/home/waishnav/.devspace/worktrees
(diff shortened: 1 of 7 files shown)


Waishnav:
Thanks for working on this. We chose not to merge this approach because it added a separate JSON persistence path beside the existing SQLite state and kept access tokens in memory. DevSpace v1.0.2 now fixes restart persistence using the existing SQLite database, with automatic versioned migrations, transactional refresh-token rotation, replay protection, and persisted access tokens. Closing this as superseded by v1.0.2.

PR #21, opened 2026-06-20
Persist OAuth clients and refresh tokens

8866Jason:
Fixes #2.

This persists OAuth registered clients so MCP clients do not hit `invalid_client` after the DevSpace server restarts.

It also persists refresh tokens using hashed token keys only. Persisting clients alone still leaves existing sessions unable to refresh after restart.

Validation:
- `npm run typecheck`
- `npm test`
- `npm run build`

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "build:app": "vite build",
     "dev": "node scripts/dev-server.mjs",
     "start": "node dist/cli.js serve",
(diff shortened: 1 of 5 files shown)


Waishnav:
Thanks for the contribution and for covering the restart case. We went with the SQLite implementation in DevSpace v1.0.2 instead of separate client and refresh-token JSON files. This keeps all state under one migration system and makes refresh-token consumption and replacement transactional, including concurrent replay protection. Closing this as superseded by v1.0.2.

PR #23, opened 2026-06-20
Persist OAuth state in SQLite

alvarolordelo:
Persist dynamically registered OAuth clients, authorization codes, access tokens, and refresh tokens in the existing SQLite state database.

Store authorization codes and tokens as SHA-256 hashes instead of raw values, preserve OAuth scopes and resources, and remove expired records.

Add graceful SQLite shutdown, UTF-8 BOM handling for configuration files, and tests covering persistence, expiration, revocation, and token hashing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "build:app": "vite build",
     "dev": "node scripts/dev-server.mjs",
     "start": "node dist/cli.js serve",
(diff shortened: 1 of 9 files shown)


Waishnav:
Thanks for the SQLite implementation. DevSpace v1.0.2 has now shipped the fix through a separate implementation with centralized versioned migrations, atomic refresh-token rotation, concurrent replay protection, stricter state-file permissions, and expanded restart/security tests. Closing this as superseded by v1.0.2.

PR #27, opened 2026-06-22
[codex] Persist OAuth state and improve network setup

Lihatoo:
## What changed

- add persistent bind-host configuration and document direct network binding
- add Localhost vs Custom URL selection to `devspace init`
- preserve CLI executable permissions after builds
- persist dynamically registered OAuth clients plus hashed access/refresh tokens in SQLite
- add a restart-persistence test for OAuth state
- document `/mcp` URL requirements and `invalid_client` recovery

## Why

ChatGPT reuses the `client_id` returned by dynamic client registration. DevSpace previously stored registered clients and tokens only in memory, so any server restart caused the reused client ID to fail with `invalid_client`. Users then had to delete and recreate the ChatGPT app.

The network setup also required manual workarounds when binding outside `127.0.0.1`, and setup always required a custom public URL.

## Impact

After one final app reconnection when upgrading from the old in-memory implementation, OAuth client IDs and refresh state survive DevSpace restarts. Operators can bind directly to a selected interface and choose localhost during setup when no public URL is needed.

Only token hashes are stored in SQLite; raw access and refresh tokens are not persisted.

## Validation

- `npm run typecheck`
- `npm test`
- `npm run build`
- registered a DCR client, restarted DevSpace, and verified the same `client_id` still opened `/authorize` with HTTP 200
- verified local and public `/healthz` endpoints return HTTP 200

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 node_modules/
 dist/
 releases/
+workspace/
 .env
 *.log
--- a/README.md
+++ b/README.md
@@ -48,13 +48,16 @@ During setup, DevSpace asks for:
 
 - the local project folders ChatGPT is allowed to open through DevSpace
 - the local port, usually `7676`
-- whether clients connect through localhost or a custom URL
+- the local listen address, such as `127.0.0.1`, `100.64.0.2`, or `0.0.0.0`
+- whether clients connect through localhost or a custom HTTPS URL
 
 Choose **Localhost** to use `http://localhost:<port>` automatically. Choose
-**Custom URL** to enter a Cloudflare Tunnel, ngrok, Pinggy, Tailscale Funnel,
+**Custom HTTPS URL** to enter a Cloudflare Tunnel, ngrok, Pinggy, Tailscale Funnel,
 reverse proxy, or other network origin.
 
-When choosing **Custom URL**, enter the public origin without `/mcp`:
+When an HTTPS URL already exists, `devspace init --force` defaults to keeping it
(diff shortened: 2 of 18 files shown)


Waishnav:
Thanks for putting this together — the problem description is spot on: ChatGPT reuses DCR `client_id`s, and in-memory OAuth state caused **`invalid_client` after every `devspace serve` restart**.

**Closing without merge:** that core issue is **already fixed on `main` and rolled out in the latest published DevSpace** (currently **v1.0.3** on npm: `@waishnav/devspace`). Persistence lives in SQLite for registered clients and **hashed** access/refresh tokens, with versioned migrations (`src/oauth-store.ts` / `src/db/`) and restart + rotation coverage in `src/oauth-store.test.ts`. This branch was based on an older point in history; merging it now would **reintroduce a parallel OAuth implementation** (inline tables in `oauth-provider`, dropping `migrateDatabase` in `db/client`) and conflict with the oauth-persistence hardening that already shipped.

**Still useful from this PR** (we may pull in separately on current `main`): bind-host prompt in `devspace init`, Localhost vs Custom HTTPS URL, `devspace config set host`, OAuth/network docs (`oauth-state-and-security.md`, `invalid_client` recovery), OAuth request path logging, and `chmod` on `dist/cli.js` after build.

Appreciate the validation notes and the one-time “re-add ChatGPT app after upgrade” guidance — worth keeping that documented for anyone upgrading from pre-persistence builds, even though new installs on the latest release get persistence out of the box.

PR #40, opened 2026-06-25
[codex] Fix Windows allowedRoots bypass

### LINKED ISSUE #13: 'allowedRoots' have some problems
No matter how I set the parameter, Chagpt Web can only access _C:\Users\Administrator_ and can even perform operations under that document, which is extremely dangerous.

Image

Image

Ahmed-Hindy:
## Summary

Fixes #13.

This PR closes a Windows-specific `allowedRoots` bypass where a path on a different drive from the configured allowlist root could be treated as allowed.

## Root cause

`isPathInsideRoot()` uses `path.relative(resolvedRoot, resolvedPath)` to decide whether a requested path is contained by an allowed root. The existing check rejected relative paths that escaped the root with `..` segments:

```ts
!relationship.startsWith("..") &&
relationship !== ".." &&
!relationship.includes(`..${sep}`)
```

That is enough for same-drive Windows paths and POSIX paths, but it misses an important Windows behavior: when `path.relative()` compares paths on different drives, it can return an absolute path instead of a `..`-prefixed relative path.

For example:

```ts
path.relative(
  "G:\\Projects\\Dev\\Github\\devspace",
  "C:\\Users\\Administrator",
);

// "C:\\Users\\Administrator"
```

Because `"C:\\Users\\Administrator"` does not start with `..`, the old predicate could incorrectly accept it as inside the allowed root.

## Security impact

This undermines the filesystem allowlist boundary on Windows. A server configured with a narrow allowlist such as:

```txt
G:\Projects\Dev\Github\devspace
```

could still allow MCP clients to open a workspace under a different drive, such as:

```txt
C:\Users\Administrator
```

That matches the class of behavior reported in #13 and can allow reads/writes outside the configured approved roots.

## Changes

- Import `isAbsolute` from `node:path`.
- Reject absolute `path.relative()` results inside `isPathInsideRoot()`.
- Add a Windows-only regression test that verifies `C:\Users\Administrator` is rejected when the only allowed root is on `G:\...`.

## Why this fix

A valid contained path should produce either:

- an empty relationship, when the path is the root itself; or
- a non-absolute relative path that does not escape through `..`.

So the containment check now requires the relationship to be non-absolute before accepting it:

```ts
!isAbsolute(relationship)
```

This preserves the existing same-drive behavior while closing the cross-drive Windows bypass.

## Validation

Automated checks run on Windows:

```txt
npx tsx src/roots.test.ts
npm run typecheck
npm test
npm run build
```

All passed.

Manual QA was also performed on Windows through ChatGPT / DevSpace Local5 with the fixed build running.

DevSpace was restarted with this configured allowed root:

```txt
G:\Projects\Dev\Github\devspace-test
```

Confirmed through ChatGPT / DevSpace Local5 that the configured root is accessible:

```txt
G:\Projects\Dev\Github\devspace-test
```

Confirmed that the previous DevSpace source folder is no longer accessible after restart:

```txt
G:\Projects\Dev\Github\devspace-1.0.2
```

Confirmed that unrelated paths on other drives are rejected as outside allowed roots:

```txt
C:\Users\Ahmed Hindy\nuke-aov-rebuilder
F:\Users\Ahmed Hindy\Documents\houdiniTools\houdini_asset_relinker
```

Also confirmed broader parent/sibling probes are blocked, so ChatGPT cannot discover access by scanning upward outside the configured root.

Manual QA result:

- configured root opens successfully
- previous/sibling project root is blocked
- unrelated `C:\...` and `F:\...` paths are blocked
- broader parent paths are blocked
- Windows cross-drive allowlist bypass appears fixed

Additional observation: DevSpace does not currently expose a tool that lists configured allowed roots directly to the MCP client. ChatGPT can only verify access by attempting to open a specific absolute path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved allowed-path checking to better handle Windows and other edge cases where path comparisons can be misleading.
  * Tightened path validation so paths outside the permitted roots are more reliably rejected.
  * Added a Windows-specific test to verify invalid home directory paths are blocked with the expected error message.

### DIFF
--- a/src/roots.test.ts
+++ b/src/roots.test.ts
@@ -24,3 +24,10 @@ assert.equal(
   resolveAllowedPath("~/file.txt", "/workspace", ["/workspace"]),
   resolve("/workspace", "~/file.txt"),
 );
+
+if (process.platform === "win32") {
+  assert.throws(
+    () => assertAllowedPath("C:\\Users\\Administrator", ["G:\\Projects\\Dev\\Github\\devspace"]),
+    /Path is outside allowed roots/,
+  );
+}
--- a/src/roots.ts
+++ b/src/roots.ts
@@ -1,5 +1,5 @@
 import { homedir } from "node:os";
-import { relative, resolve, sep } from "node:path";
+import { isAbsolute, relative, resolve, sep } from "node:path";
 
 export class AccessDeniedError extends Error {
   constructor(message: string) {
@@ -24,7 +24,10 @@ export function isPathInsideRoot(path: string, root: string): boolean {
 
   return (
     relationship === "" ||
-    (!relationship.startsWith("..") && relationship !== ".." && !relationship.includes(`..${sep}`))
+    (!isAbsolute(relationship) &&
+      !relationship.startsWith("..") &&
+      relationship !== ".." &&
+      !relationship.includes(`..${sep}`))
   );
 }
 


Ahmed-Hindy:
**Potential Issues**
PR #40 fixes the Windows cross-drive bug, but a few things could still go wrong:

1. **Junctions or symlinks can still escape the allowed folder**

Example:

```txt
Allowed root:
G:\Projects\SafeFolder

Inside it:
G:\Projects\SafeFolder\escape

But escape is a junction pointing to:
C:\Users\Ahmed\Secrets
```

DevSpace may think this is allowed because the typed path starts inside `SafeFolder`, even though Windows resolves it somewhere else.

This is the most important remaining risk.

2. **Some valid Windows path forms may be rejected**

Example:

```txt
C:\repo\file.txt
\\?\C:\repo\file.txt
```

These can refer to the same place, but the current check may treat them as different and reject the second one.

That is safer than allowing too much, but it may annoy users with unusual Windows path formats.

3. **Mapped drives and network shares may behave unexpectedly**

Example:

```txt
Z:\project
\\server\share\project
```

These might point to the same folder, but the current check may not know that unless it resolves the real filesystem path.

Likely result: false rejection, not a security bypass.

4. **Test coverage is still narrow**

The PR only adds one Windows regression test:

```txt
G:\allowed-root
C:\outside-root
```

That proves the reported bug is fixed, but it does not cover junctions, symlinks, network paths, mapped drives, or weird Windows path formats.

5. **Some clients may lose access they previously had**

If someone was accidentally relying on the old broken behavior, this PR will block them.

Example:

```txt
Allowed root:
G:\Projects

Previously accepted by mistake:
C:\Users\Ahmed
```

That is a good break, because the old behavior was unsafe, but it is still a behavior change.

**Future PR**
I would keep PR #40 as the small direct fix, then open a separate hardening PR.

That future PR should:

1. **Add real filesystem canonicalization**

Use `fs.realpath()` or similar so DevSpace compares where paths actually point, not just what the string looks like.

2. **Block symlink and junction escapes**

If a path inside an allowed root points outside through a junction/symlink, reject it.

3. **Handle new files carefully**

For reads, the file exists, so `realpath()` is easy.

For writes, the file may not exist yet. In that case, resolve the nearest existing parent folder and make sure that parent is still inside the allowed root.

4. **Add more Windows tests**

Add tests for:

```txt
junction inside allowed root pointing outside
symlink inside allowed root pointing outside
same root path
child path
sibling path
C:\ vs G:\ cross-drive path
UNC paths like \\server\share
mapped-drive style paths
\\?\C:\... namespaced paths
mixed-case paths
trailing slash paths
```

5. **Document the expected behavior**

The security docs should say that allowed roots are resolved to their real filesystem locations, and that symlinks/junctions cannot be used to escape them.

So: PR #40 is good and fixes the reported bug. The next PR should be a broader “canonical path containment” security hardening pass.

Waishnav:
Thanks for contribution @Ahmed-Hindy, will be live in new release, Hopefully by end of today :)

PR #43, opened 2026-06-27
Fix checkout workspace opening for Windows drive roots

dodjdnh:
Fixes Windows drive-root checkout workspace initialization.

This fixes a Windows-specific checkout workspace issue where opening an existing drive root such as `D:\` or `E:\` can fail with:

`EPERM: operation not permitted, mkdir 'D:\'`

Root cause:

`openCheckoutWorkspace()` called `mkdir(root, { recursive: true })` before checking whether the path already existed. That is unnecessary for existing paths and can fail for Windows drive roots.

What changed:

* Added `ensureCheckoutWorkspaceRoot()`
* `stat()` is now attempted first
* `mkdir(..., { recursive: true })` is only called when `stat()` fails with `ENOENT`
* Existing behavior is preserved for missing checkout workspace directories
* Added a regression test to ensure existing directories do not call `mkdir()`

Validation:

* `npm run test`
* `npm run typecheck`
* `npm run build`

Manual QA:

A short Windows manual QA recording will be attached in the PR discussion.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workspace checkout handling so the root directory is verified consistently before use.
  * Existing workspace roots are now detected without creating unnecessary directories.
  * Added coverage to confirm no filesystem changes occur when the checkout root already exists.

### DIFF
--- a/src/workspaces.test.ts
+++ b/src/workspaces.test.ts
@@ -7,7 +7,7 @@ import assert from "node:assert/strict";
 import { loadConfig } from "./config.js";
 import { GitWorktreeError } from "./git-worktrees.js";
 import { SqliteWorkspaceStore } from "./workspace-store.js";
-import { WorkspaceRegistry } from "./workspaces.js";
+import { ensureCheckoutWorkspaceRoot, WorkspaceRegistry } from "./workspaces.js";
 
 const execFileAsync = promisify(execFile);
 const root = await mkdtemp(join(tmpdir(), "devspace-workspace-test-"));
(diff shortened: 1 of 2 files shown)


dodjdnh:
https://github.com/user-attachments/assets/dab0b764-d033-441a-8648-112c58e41b9f

dodjdnh:
Manual QA recording attached. Local Windows testing for the drive-root checkout fix:

* Branch: `fix/windows-drive-root-checkout`
* Original failing operation: `mkdir("D:\\", { recursive: true })` fails with `EPERM`
* Original failing operation: `mkdir("E:\\", { recursive: true })` fails with `EPERM`
* Fixed DevSpace code path: `ensureCheckoutWorkspaceRoot("D:\\")` succeeds and returns `isDirectory = true`
* Fixed DevSpace code path: `ensureCheckoutWorkspaceRoot("E:\\")` succeeds and returns `isDirectory = true`
* `npm run build` passes
* `node dist/cli.js doctor` shows the temporary QA allowed roots, including `D:\`, `E:\`, and `E:\MinerU`

I also tried running the full ChatGPT + DevSpace `open_workspace` flow against `D:\` / `E:\`, but the tool call was blocked by OpenAI's safety check before it reached DevSpace. That is a platform-level restriction, not the DevSpace bug, so I verified the Windows filesystem behavior locally against the fixed DevSpace code path instead.

Waishnav:
Thanks @dodjdnh for the clean fix here, and for including both automated coverage and Windows manual QA. I’ve merged this.

PR #67, opened 2026-07-08
Add Executor local MCP skill example

JGSphaela:
## Summary

Reworks the Executor experiment away from DevSpace bridge tools and into a skill-based workflow.

This PR now adds an example Agent Skill at `examples/skills/executor-local-mcp/SKILL.md` that teaches the model to use the user's existing `executor` CLI through DevSpace's existing shell tools (`bash` or `exec_command`) to:

- inspect configured Executor sources
- search the Executor tool catalog
- describe unfamiliar tools before calling them
- invoke local MCP/API tools through `executor call`

It also documents how to copy the example skill into `~/.devspace/skills` or another active skill directory.

## Why this direction

The earlier bridge-tool approach worked as a proof of concept, but it duplicated what the local Executor CLI can already do from shell tools and added more model-facing tool names.

I also checked `@executor-js/sdk`; from its package docs/types it looks aimed at building or embedding Executor runtimes/plugins rather than acting as a thin client for a user's already-running local Executor service. So for this iteration, the skill approach keeps DevSpace core unchanged and avoids wrapping the CLI.

## Validation

- `npm run typecheck`
- `npm test`
- `npm run build`
- Local Executor CLI smoke test:
  - `executor tools sources`
  - `executor tools describe zotero.user.default.zotero_list_collections`
  - `executor call zotero user default zotero_list_collections '{}'`


Waishnav:
tbh, that’s not how I would approach it. find out is there an executor SDK we can use for this implementation?

The current approach doesn’t really make sense to me because:

1. It adds extra tool calls and basically just wraps the local executor CLI.
2. The executor can already be called from the `bash` or `exec_command` tool, so we probably just need to educate the model through a custom skill for external MCP tools available through executor.

JGSphaela:
That's fair, it just something I did for proof of concept and to get it working. I didn't know that there's a SDK already, I'll check it out!

JGSphaela:
I scrapped the bridge-tool approach.

I checked `@executor-js/sdk`; from the package docs/types it looks more aimed at building or embedding Executor runtimes/plugins than acting as a thin client for a user's already-running local Executor service.

I pushed a new revision that keeps DevSpace core unchanged and adds an example skill instead. The skill teaches the model to use the existing `executor` CLI through `bash`/`exec_command` for local MCP/API integrations.

PR #71, opened 2026-07-11
fix: clean up stale MCP sessions to bound memory growth

Zhenyu98:
## Summary

- add an `McpSessionRegistry` that tracks the last activity time for each Streamable HTTP transport
- close sessions after 24 hours of inactivity with a conservative five-minute cleanup sweep
- close and release remaining transports during server shutdown
- log whether a session closed normally, timed out, or failed to close
- add regression coverage for activity refresh, idle cleanup, close failures, explicit removal, and shutdown cleanup

## Motivation

`createServer()` currently keeps every initialized MCP transport in a process-lifetime `Map` and removes it only when `transport.onclose` fires. Some MCP clients reconnect without explicitly closing the previous transport, so abandoned transports and their connected MCP server/tool objects can remain strongly referenced indefinitely.

In one long-running deployment, the logs contained 96 `mcp_session_created` events and zero `mcp_session_closed` events. The DevSpace process grew from about 218 MB of private memory immediately after a restart to about 2.55 GB after roughly two weeks, then returned to the baseline after restarting. This observation does not prove that every retained byte came from MCP transports, but it demonstrates the unbounded-retention risk and motivated this memory-usage fix.

## Behavior

- every request using an existing session refreshes its last-activity timestamp
- sessions inactive for 24 hours are removed from the registry before `transport.close()` is awaited, so references are released even when closing fails
- the 24-hour timeout is intentionally conservative to avoid disrupting normal pauses in a ChatGPT or Claude conversation
- a client that attempts to reuse an expired session receives the existing `Unknown MCP session` response and can initialize a new session
- normal `transport.onclose` handling remains supported and no duplicate close log is emitted

## Verification

- `npm test`
- `npm run typecheck`
- `npm run build`
- `git diff upstream/main...HEAD --check`

The change is intentionally limited to MCP transport lifecycle management. Workspace-cache eviction can be evaluated separately.

## Summary to CodeRabbit

- **New Features**
  - Added an in-memory MCP session registry with idle-timeout cleanup and bulk shutdown.
  - Improved graceful shutdown by draining the HTTP server, awaiting async app cleanup, and reporting per-transport close outcomes.

- **Bug Fixes**
  - Ensured session removal is respected during shutdown (removed sessions are no longer closed).

- **Tests**
  - Added coverage for idle cleanup, transport close failures, registry shutdown semantics, delayed HTTP draining, and HTTP close error propagation.
  - Expanded the test runner to include the new shutdown and session tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "dev": "node scripts/dev-server.mjs",
     "postinstall": "node scripts/fix-node-pty-permissions.mjs",
     "start": "node dist/cli.js serve",
-    "test": "tsx src/config.test.ts && tsx src/ui/card-types.test.ts && tsx src/ui/patch-display.test.ts && tsx src/ui/tool-display.test.ts && tsx src/apply-patch.test.ts && tsx src/process-platform.test.ts && tsx src/process-sessions.test.ts && tsx src/local-agent-runtime.test.ts && tsx src/local-agent-adapters.test.ts && tsx src/local-agent-availability.test.ts && tsx src/local-agent-profiles.test.ts && tsx src/local-agent-targets.test.ts && tsx src/local-agent-store.test.ts && tsx src/roots.test.ts && tsx src/skills.test.ts && tsx src/workspaces.test.ts && tsx src/review-checkpoints.test.ts && tsx src/oauth-store.test.ts && tsx src/cli.test.ts",
+    "test": "tsx src/config.test.ts && tsx src/ui/card-types.test.ts && tsx src/ui/patch-display.test.ts && tsx src/ui/tool-display.test.ts && tsx src/apply-patch.test.ts && tsx src/process-platform.test.ts && tsx src/process-sessions.test.ts && tsx src/mcp-sessions.test.ts && tsx src/local-agent-runtime.test.ts && tsx src/local-agent-adapters.test.ts && tsx src/local-agent-availability.test.ts && tsx src/local-agent-profiles.test.ts && tsx src/local-agent-targets.test.ts && tsx src/local-agent-store.test.ts && tsx src/roots.test.ts && tsx src/skills.test.ts && tsx src/workspaces.test.ts && tsx src/review-checkpoints.test.ts && tsx src/oauth-store.test.ts && tsx src/cli.test.ts",
     "typecheck": "tsc -p tsconfig.json --noEmit"
   },
   "keywords": [],
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "dev": "node scripts/dev-server.mjs",
     "postinstall": "node scripts/fix-node-pty-permissions.mjs",
     "start": "node dist/cli.js serve",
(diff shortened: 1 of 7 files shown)


Zhenyu98:
@coderabbitai review

Waishnav:
Thanks for the PR @Zhenyu98  — the idle-session cleanup is a meaningful RAM-usage improvement. The current session map can retain abandoned transports and their connected MCP server object graphs indefinitely, so bounding that lifecycle is a valuable fix for long-running DevSpace processes.

The registry implementation itself looks sound, but the server-shutdown cleanup is currently ordered in a way that can prevent it from running.

Both entrypoints call `close()` only inside the callback passed to `httpServer.close()`. Node waits for active HTTP requests to finish before invoking that callback. An MCP SSE request can remain active until its `StreamableHTTPServerTransport` is closed, but those transports are only closed by `close()`. This creates a circular wait: HTTP shutdown waits for the SSE request, while transport shutdown waits for the HTTP shutdown callback.

There is also a second race: `close()` starts `transports.closeAll()` without awaiting it, and the caller immediately invokes `process.exit(0)`. That can terminate the process before transports finish closing or shutdown logs are emitted.

Could we make `RunningServer.close()` return `Promise<void>` and await `transports.closeAll()`? The signal handler can start `httpServer.close()` and then await application cleanup while the HTTP server is draining:

```ts
const shutdown = async () => {
  const httpClosed = new Promise<void>((resolve, reject) => {
    httpServer.close((error) => {
      if (error) reject(error);
      else resolve();
    });
  });

  await close();
  await httpClosed;
};
```

This lets closing the MCP transports terminate active SSE responses, which in turn allows `httpServer.close()` to complete. It also avoids relying on asynchronous cleanup after `process.exit()`.

Zhenyu98:
Thanks for digging into this — that was a sharp catch. @Waishnav 

The original change was focused on bounding the lifetime of abandoned MCP sessions. Your shutdown analysis connected that fix to the server’s termination path and exposed two issues that had been missed: the circular wait between `httpServer.close()` and an active SSE transport, and the race between asynchronous cleanup and `process.exit(0)`.

Both cases were reproduced, and the implementation was updated in `ee372e7`. `RunningServer.close()` now returns a memoized `Promise<void>` and waits for all MCP transports to close. In both entrypoints, HTTP draining is started first, application cleanup runs immediately, and the process exits only after both cleanup and HTTP drain have completed.

Focused regression tests were added for shutdown ordering, delayed transport closure, HTTP drain completion, and HTTP close failures. The branch was also rebased onto the current `main` (`b027795`). The full test suite, typecheck, production build, and `git diff --check` all pass.

Waishnav:
@Zhenyu98 Thanks for the PR and improvement it is bringing to devspace :)

whitesarum:
Additional production evidence from a high-churn ChatGPT workload:

- Environment: macOS, Node 24.18.0, DevSpace stateful Streamable HTTP session registry.
- Collector baseline: 144 retained MCP sessions, heapUsed 155,141,692 B, physical footprint 229,591,480 B.
- After ~2h12m: 586 sessions, heapUsed 442,791,944 B. The run stopped on a +256 MiB heap safety threshold.
- Longer advisory observation before restart: 3,904 current / 4,134 peak sessions, heapUsed 2,855,774,792 B, physical footprint 3,269,620,040 B.
- The 24h cleanup did run: 3,301 idle_timeout removals and 3,301 close_completed, with 0 close_failed. Despite that, sustained creation still left ~3.9k sessions retained.
- Session count and heapUsed moved together strongly (Pearson ~0.998 in the collector window, ~0.993 over the longer observation).
- Restart returned process footprint to roughly 169 MB.

So #71 is definitely addressing a real retention path, and the cleanup itself appears to work. The additional finding is that a 24h horizon is not a deterministic memory bound under a much higher ChatGPT reconnect/initialize rate (observed roughly 2.5–3.4 new sessions/minute).

I am not claiming retained sessions are the sole cause of every timeout or disconnect. This is specifically evidence that the current 24h policy can be outpaced by high-churn clients and can still permit multi-GB growth.

PR #84, opened 2026-07-18
feat: add MCP client access policy

Zhenyu98:
## Summary

- add an opt-in MCP client denylist with `off` and `enforce` modes
- inspect `initialize.params.clientInfo` before creating the MCP transport
- reject matching clients with HTTP 403 and JSON-RPC error `-32003`
- add structured client observation and denial logs
- add regression coverage for Codex detection and clients without `clientInfo`

## Motivation

ChatGPT and Codex currently share connected apps. DevSpace is intended to let ChatGPT control the local PC remotely, while Codex already runs locally and should use its native filesystem, shell, Git, and patch tools.

Without a server-side guard, Codex may accidentally invoke DevSpace instead of its native tools.

## Behavior

Enable the guard with:

```env
DEVSPACE_CLIENT_ACCESS_MODE=enforce
DEVSPACE_DENIED_CLIENTS=codex
````

Clients matching the denylist are rejected before an MCP transport or workspace session is created. Clients that omit `clientInfo` are allowed to preserve current ChatGPT compatibility.

Practical deployment testing confirmed:

* ChatGPT `open_workspace` succeeded
* Codex `open_workspace` returned HTTP 403, JSON-RPC `-32003`, and `MCP client not allowed`
* Codex MCP App resource reads were also denied

`clientInfo` is self-declared, so this is a guard against accidental client use rather than cryptographic isolation.

## Verification

* `npm test`
* `npm run typecheck`
* `npm run build`
* `git diff upstream/main...HEAD --check`
* live ChatGPT connection test
* live Codex denial test

## Summary by CodeRabbit

- **New Features**
  - Added optional client access controls for MCP initialization requests.
  - Supports deny-listed client identities and returns HTTP 403 with a clear JSON-RPC error when access is denied.
- **Configuration**
  - Added environment and configuration-file options for access mode and denied clients.
  - Access controls remain disabled by default unless enforcement is enabled.
- **Tests**
  - Added coverage for client identification, access decisions, configuration validation, and HTTP behavior for allowed and denied requests.


Waishnav:
isn't there is way to disable MCP in codex? i.e you can disable the `devspace` plugin for local instance ig

Zhenyu98:
@Waishnav Thanks. Codex does expose a disable control, but DevSpace is inherited as a shared ChatGPT app rather than configured as a local [mcp_servers.devspace] entry.

I tested disabling DevSpace from Codex’s Apps/Plugins UI. That change is synchronized to the shared app state and also removes/disables DevSpace in ChatGPT.

OpenAI’s documentation also notes that app settings are shared and disabling an app can affect other ChatGPT experiences using the same app.

The desired setup is:

keep DevSpace connected and available in ChatGPT;
prevent only Codex from invoking it;
let Codex use its native local filesystem, shell, Git, and patch tools.

Since there is currently no Codex-local entry for this shared app in config.toml, the server-side client policy is intended to provide that separation.

PR #88, opened 2026-07-20
feat: download native ChatGPT files into workspaces

jplew:
## Summary

Replace the artifact-service lifecycle with one one-shot handoff:

```text
open_workspace
download_artifact({ file, workspaceId }) -> { path }
```

DevSpace validates the native MCP-host file reference, streams it directly into the selected workspace under `.devspace/incoming/`, chooses a normalized collision-free filename, and returns only the workspace-relative path.

## Tool contract

- Exactly one artifact-specific tool: `download_artifact`.
- Inputs are only the native `file` value and an existing `workspaceId`.
- The descriptor retains `openai/fileParams: ["file"]`.
- Output is only `{ path }`.
- Destination selection, conflict policy, expected hashes, artifact IDs, host paths, TTLs, pinning, stat/delete/copy, and upload lifecycle are not model-facing.

## Implementation

- Streams under `DEVSPACE_ARTIFACT_MAX_FILE_BYTES` while computing SHA-256.
- Publishes into `.devspace/incoming/` with deterministic collision suffixes.
- Uses descriptor-anchored directory creation and publication so pathname replacement cannot redirect writes outside the workspace.
- Chmods, fsyncs, and verifies through the still-open partial file descriptor; no path-based chmod or hash occurs after publication.
- Performs bounded cleanup of owned stale partials only.
- Preserves the trusted OpenAI native-file adapter, regional host restrictions, redirect validation, and log redaction.

## Removed lifecycle

- No persistent artifact object store or completed-object database.
- No upload/chunk protocol, artifact IDs, quota ledger, TTL, pinning, background cleanup timer, or artifact root.
- Fresh databases do not create artifact tables.

## Migration behavior

Existing legacy artifact tables and bytes are intentionally left untouched. The new implementation does not read, mutate, or opportunistically delete prior user data; cleanup can be performed manually after review.

## Verification

- `npm test`
- `npm run typecheck`
- `npm run build`
- `git diff --check`

No runtime deployment or service restart was performed as part of this source-only handoff.

## Summary by CodeRabbit

* **New Features**
  * Added optional native file downloads directly into an open workspace.
  * Downloads are limited by configurable file-size settings and cannot overwrite existing files.
  * Support is enabled with `DEVSPACE_ARTIFACTS=1` and is available on supported non-Windows platforms.
  * Added secure validation for file sources, destinations, redirects, and sensitive data handling.

* **Documentation**
  * Added setup, usage, configuration, security, and logging guidance for native file downloads.

* **Tests**
  * Added coverage for validation, size limits, conflicts, path safety, cleanup, permissions, and redacted logging.


jplew:
Addressed the initial review findings in a single amended commit:

- replaced the personal artifact-root example with `$HOME/...`
- centralized and enforced the maximum configured artifact TTL, with an upper-bound regression test
- aligned Drizzle index definitions with the migration's descending order using `desc(...)`

Re-ran `npm test`, `npm run typecheck`, `npm run build`, and `git diff --check upstream/main...HEAD`; all passed.

package.json
```diff
@@ -28,7 +28,7 @@
...
+    "test": "tsx src/config.test.ts && tsx src/artifacts.test.ts && tsx src/incoming-artifacts.test.ts && tsx src/artifact-workspace.test.ts && tsx src/ui/card-types.test.ts && tsx src/ui/patch-display.test.ts && tsx src/ui/tool-display.test.ts && tsx src/apply-patch.test.ts && tsx src/process-platform.test.ts && tsx src/process-sessions.test.ts && tsx src/mcp-sessions.test.ts && tsx src/server-shutdown.test.ts && tsx src/local-agent-runtime.test.ts && tsx src/local-agent-adapters.test.ts && tsx src/local-agent-availability.test.ts && tsx src/local-agent-profiles.test.ts && tsx src/local-agent-store.test.ts && tsx src/roots.test.ts && tsx src/skills.test.ts && tsx src/workspaces.test.ts && tsx src/review-checkpoints.test.ts && tsx src/oauth-store.test.ts && tsx src/cli.test.ts",
```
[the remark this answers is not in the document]
jplew: Addressed in 5edc5a4: restored src/local-agent-targets.test.ts to the npm test chain, and expanded artifact workspace coverage for conflict handling, path escape rejection, and symlink rejection.

src/artifact-workspace.ts
```diff
@@ -0,0 +1,231 @@
...
+      size: finalEntry.size,
+      sha256: `sha256:${digest}`,
+      onConflict: input.onConflict,
+      renamed: finalPath !== input.destination,
+    };
+  } catch (error) {
+    await unlink(tempPath).catch(() => undefined);
+    throw error;
+  }
+}
```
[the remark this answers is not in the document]
jplew: Addressed in 5edc5a4: materialized destination cleanup now removes only a path created by this operation after a downstream verification/permission failure; the temporary path is still cleaned independently.

Waishnav:
@jplew Really cool feature and PR

I'ven't gone through whole changes yet but these are my first impressions
1. why so many tool call expose to model? isn't there should be just single tool call like `download_artifact`? like life-cycle around around artifact files are even needed at all, if those files can be moved/delete/renamed using other tools like `bash` or `exec_command`
2. can't we handle this on tooling level only like by default generated artifact will be auto-downloaded inside `~/.local/share/devspace/artifact/` i.e no need to expose this to model

philosophy of devspace usually is to keep things very minimally expose to model as much as possible and handle at tooling level if it's possible

will test it by tomorrow but yeah pretty cool idea which i never thought could be possible (it seems like chatgpt apps sdk are pretty powerful)

jplew:
> philosophy of devspace usually is to keep things very minimally expose to model as much as possible and handle at tooling level if it's possible

Thanks for reviewing, all good points. I can definitely abstract away internals and consolidate everything into a single call. Will post a follow-up soon.

jplew:
done, condensed everything into a single call.
image

Waishnav (changes requested):
Thanks for consolidating this into one model-facing call — that direction matches DevSpace well. I ran the full test suite, typecheck, build, and `git diff --check`; they pass.

Before merging, I think this should be trimmed to a one-shot handoff rather than a persistent artifact service:

```ts
download_artifact({ file, workspaceId }) -> { path }
```

DevSpace can choose a collision-free path under `.devspace/incoming/`; existing filesystem tools can move/rename/delete it. This would let us remove model-facing `destination`, `onConflict`, and `expectedSha256`, plus the internal upload/chunk/TTL/pinning/quota/database lifecycle.

There are also four filesystem-safety blockers in the current implementation:

1. `ensureSecureDirectory()` creates/chmods before proving containment, so an intermediate symlink can cause changes outside the artifact root.
2. Workspace materialization performs path-based chmod/hash/cleanup after rename, leaving a TOCTOU window where the destination can be swapped.
3. Artifact deletion removes the DB row before successfully removing the object; failures leave permanent untracked bytes and incorrect quota/cleanup state.
4. Startup unconditionally chmods an existing configured artifact root to `0700`, which can silently break shared or service-managed directories.

Please simplify the implementation and address these before merge. The native-file adapter, streaming validation, size enforcement, safe filename handling, and tests are all valuable and can remain in the smaller design.

jplew:
@Waishnav I made the requested change. Test looks like this so it appears to be working: 
image

Last message in attached convo is an easy move.

Waishnav:
@jplew heyo can you retest real quick, i've added back the `path` param and removed that `.devspace/incoming/` as default download folder i.e letting model decide where it wants to download it, just want to make sure everything is all right with my recent changes

jplew:
> @jplew heyo can you retest real quick, i've added back the `path` param and removed that `.devspace/incoming/` as default download folder i.e letting model decide where it wants to download it, just want to make sure everything is all right with my recent changes 

Yes will retest your changes after a few hours

jplew:
@Waishnav Yes, I tested on the latest commit (45fb361) and it's still working as expected, see screenshot here: https://media.haft.sh/dev/asset/asset-img-4215022a5cef-png/01cf6042/img-4215022a5cef.png

Tested with an image and a markdown file

Waishnav:
@coderabbitai can you do one last review before merging this

jplew:
@Waishnav want me to address these coderabbit requested changes?

docs/security.md
```diff
@@ -92,9 +92,65 @@ Managed worktrees reduce accidental edits to your active checkout, but they are
...
+alphanumeric. Arbitrary Azure Blob accounts, alternate ports, credentials,
+fragments, malformed IDs, extra fields, and redirects outside that boundary fail
+closed. Opaque IDs are bounded metadata and are never used as filenames or path
+components.
+
+`download_artifact` accepts only the native file value, a `workspaceId` returned
+by `open_workspace`, and a relative destination `path`. Absolute paths,
+traversal, symlinked parents, and existing destinations fail closed. Callers
+cannot supply a conflict mode, expected hash, host path, or storage policy.
+
```
Waishnav: make this section smaller

Waishnav (changes requested):
yeah go ahead @jplew :) tbh currently lots of test is slop but will clean it later down the line in upcoming weeks when i'll be working on test suite overhaul

docs/artifact-exchange.md
```diff
@@ -0,0 +1,96 @@
+# Native file download
```
Waishnav: docs file changes can be trimmed and re-write it from user's POV

src/artifact-tools.ts
```diff
@@ -0,0 +1,621 @@
...
+]);
+
+const openAIFileReferenceInputSchema = z.object({
+  download_url: z.string(),
+  file_id: z.string(),
+  mime_type: z.string().nullable().optional(),
+  file_name: z.string().nullable().optional(),
+  name: z.string().nullable().optional(),
+  size: z.number().int().nonnegative().nullable().optional(),
+});
```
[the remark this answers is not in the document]
jplew: Addressed in ed2702b: the native-file input now uses z.strictObject, so unknown fields are rejected. Focused regression coverage verifies a sensitive-looking extra value is absent from validation output and artifact logging metadata.

docs/security.md
jplew: Addressed in ed2702b: reduced the native file security section to concise operator-facing behavior—native host file only, trusted handling, one-shot workspace transfer, path protections, and no arbitrary URL/local-path ingestion.

docs/artifact-exchange.md
jplew: Addressed in ed2702b: rewrote the artifact download guide around the user workflow (open_workspace → download_artifact → returned path), then normal filesystem tools. It remains aligned with the current required path parameter restored on the PR head.

jplew:
> yeah go ahead @jplew :) tbh currently lots of test is slop but will clean it later down the line in upcoming weeks when i'll be working on test suite overhaul

I addressed all the requests. Or rather, I got devspace to do it, so if it looks like slop, I'm sorry but you have to take it up with your own creation. :)

Waishnav:
lol i meant it's not the problem with what you did, tests suite and test patterns was slop earlier even before your contribution. But yeah, thanks for sticking with us throughout the review process for this really cool feature :)

PR #69, opened 2026-07-10
fix: improve compatibility with current ChatGPT MCP behavior

uniplanck:
## Summary

This draft proposes a compatibility and runtime-reliability update for the practical MCP execution regression observed after the recent ChatGPT model rollout.

The issue was not limited to GPT-5.6. In the same environment, GPT-5.5 also became materially slower or stalled during multi-step DevSpace operations, despite working normally before the rollout. This PR does not make claims about undocumented OpenAI internals; it reduces eager payload size, repeated context expansion, unnecessary tool round trips, and environment-dependent command failures.

Related to #68.

## Main changes

- compact `open_workspace` payloads with bounded instruction excerpts and explicit full reads
- exact-path reads for advertised instruction files without widening filesystem access
- optional compound inspection tools and workspace-scoped approved-shell aliases
- safe standard PATH augmentation without sourcing `.zshrc`, `.zprofile`, or another login-shell file
- execution-cost diagnostics for duration, Tool calls, errors, retries, character volume, and approximate text tokens
- bounded runtime diagnosis and compatibility smoke checks integrated into the existing Bash Tool
- model-facing Tool catalog kept compact; Finder is registered as an app-only Tool
- workspace-scoped Open in Finder actions for MCP App result cards
- optional built-in agent profiles, skill matching, design-audit tools, and generic MCP App integration improvements
- configuration documentation and automated tests

## Runtime commands

The diagnostic features do not add model-facing MCP Tools. They are intercepted through the existing Bash Tool:

- `devspace-runtime diagnose [--github] [command ...]`
- `devspace-runtime smoke`
- `devspace-runtime costs`
- `devspace-runtime finder <path>`

Finder paths are validated against the active workspace root. GitHub authentication checks report only an authenticated/unauthenticated classification and do not return credentials.

## Compatibility kit

The branch includes `compatibility-kit/openai-model-compatibility-2026-07` with six ordered patches, an apply preflight, a verification script, architecture notes, and security/privacy documentation. Bundle version is `2026.07.2`.

1. compact workspace and usage foundation
2. safe tools and compound inspection
3. agents, skills, and MCP App integration
4. runtime diagnostics and execution-cost aggregation
5. integration through the existing Tool surface
6. Finder result-card action

The updater performs no commit, push, publish, deploy, or user-configuration mutation.

## Upstream-first basis

- base repository: `Waishnav/devspace`
- base branch: current upstream `main`
[...]

PR #41, opened 2026-06-26
[codex] Add native Windows shell support

yurilopes:
﻿## What changed

This PR improves DevSpace shell execution on Windows while preserving the existing Bash behavior for Linux, macOS, and users who explicitly opt into Bash.

- Adds `DEVSPACE_SHELL=auto|bash|powershell|cmd`.
- Uses native PowerShell by default on Windows instead of routing PowerShell commands through Bash/MSYS.
- Adds native `cmd.exe` support via `DEVSPACE_SHELL=cmd`.
- Keeps Pi's Bash backend for `DEVSPACE_SHELL=bash` and non-Windows defaults.
- Adds a PowerShell guardrail that blocks fragile `-match` usage with raw Windows path literals, including literals assigned to variables and then passed to `-match`.
- Updates `devspace doctor`, MCP tool guidance, docs, and tests.

## Why

When ChatGPT generated PowerShell commands through the previous Bash-backed shell, Bash/MSYS could expand PowerShell syntax such as `$_.CommandLine` before PowerShell received it. Native PowerShell execution fixes that transport issue.

A second common failure mode remained: using PowerShell `-match` for literal Windows path fragments. `-match` is regex-based, so path fragments like `\profiles` can trigger malformed regex escapes such as `\p`. The new guardrail blocks the fragile pattern and tells the model to use `.Contains()`, `-like`, or `[regex]::Escape()`.

## Impact

Windows users can run DevSpace with native PowerShell by default. Existing Bash workflows remain available with `DEVSPACE_SHELL=bash`, and Linux/macOS behavior stays on the existing Bash backend.

## Validation

- `npm test`
- `npm run typecheck`
- `npm run build`
- `git diff --check`
- Manual global install and `devspace doctor` on Windows confirmed `Shell mode: auto` resolves to native PowerShell.

## Summary by CodeRabbit

- **New Features**
  - Added configurable shell execution across platforms, including Windows native PowerShell, `cmd.exe`, and Bash-like environments.
  - Introduced `DEVSPACE_SHELL` (`auto`, `bash`, `powershell`, `cmd`), with Windows defaulting to PowerShell.
- **Bug Fixes**
  - Improved shell tool execution with clearer output and more reliable timeout handling.
  - Added protections for fragile PowerShell matching patterns involving literal Windows paths.
- **Documentation**
  - Updated platform support, configuration docs, and Windows shell troubleshooting; enhanced `doctor` to display the active shell mode and command.
- **Tests**
[...]

PR #89, opened 2026-07-20
fix: resolve MCP session count leak causing ChatGPT TimeoutError

gufyhvvyfycyddy-code:
## Root Cause

Duplicate `sessionRegistry.markActive(sessionId)` calls for existing MCP sessions, but only one `sessionRegistry.markIdle(sessionId)` call per request. This caused the `inFlight` counter to permanently remain >= 1 for every session.

After 64 sessions accumulated, new `initialize` requests (the only sessions with `inFlight=0`) were immediately evicted by the registry, causing ChatGPT to receive `TimeoutError`.

### 1. Strict markActive/markIdle pairing (server.ts)
- Removed duplicate `markActive` call (was called twice for existing sessions)
- Single `markActive` per request, tracked via `activeSessionId` variable
- `markIdle` moved to `finally` block, only called when `activeSessionId` was set
- `initialize` requests (no `markActive`) do not call `markIdle`

```typescript
let activeSessionId: string | undefined;
try {
  if (sessionId) {
    transport = sessionRegistry.get(sessionId)?.transport;
    sessionRegistry.markActive(sessionId);
    activeSessionId = sessionId;
  } else if (initializeRequest) {
    // capacity check...
    transport = new StreamableHTTPServerTransport(...);
  }
  await transport.handleRequest(req, res, req.body);
} catch (error) {
  // error handling...
} finally {
  if (activeSessionId) {
    sessionRegistry.markIdle(activeSessionId);
  }
}
```

### 2. Improved full-load behavior (mcp-session-registry.ts)
- `register()` returns `boolean` — `false` when at capacity with no evictable idle sessions
- Added `atCapacity` getter — `true` when all 64 sessions have `inFlight > 0`
- When `atCapacity`, `initialize` requests get immediate HTTP 503 instead of being registered-then-evicted
- No "30 min force-close" logic added

### 3. Regression tests (mcp-session-lifecycle.test.ts)
8 tests covering:
1. Normal request: markActive once, markIdle once, inFlight=0
2. handleRequest throws: finally still executes, inFlight=0
3. 100 consecutive ChatGPT-style sessions: count less than or equal to 64
4. 64 idle sessions: new session evicts oldest, new survives
5. 64 active sessions: atCapacity=true, register rejected
6. Full MCP flow: initialize, notifications/initialized, tools/list, open_workspace
7. 70 consecutive requests: inFlight stays 0
8. Old bug documentation: double markActive + single markIdle leaves inFlight=1

## Verification

- All 8 regression tests pass
- All 17 original tests pass (1 pre-existing Windows symlink failure excluded)
- TypeScript typecheck passes
- 7677 candidate port validation: 7/7 tests passed
- 7676 production switch: PID 17164 to 42840, sessions 64 to 0
- 75 connection stress test: max sessions=0, no leak
- /mcp response time: 70ms (immediate, no timeout)

## Custom Windows Stable Build

Also includes previous PRs: #71 session cleanup/graceful shutdown, #43 Windows drive root fix, #41 native PowerShell, #85 12000 char output limit, #76 context scan ignore, #65/#66 AGENTS/CLAUDE realpath, #69 ChatGPT MCP compatibility.

## Summary by CodeRabbit

* **New Features**
  * Added advanced workspace, search, job-management, and Git workflow tools.
  * Added configurable shell selection, including PowerShell, Bash, and Command Prompt.
  * Added runtime health, diagnostics, smoke-test, and cost-reporting endpoints.
  * Added configurable inline output truncation with metadata.
  * Added context-ignore paths for workspace discovery.
* **Bug Fixes**
  * Improved MCP session capacity, cleanup, timeout, and close-failure handling.
  * Improved OAuth validation and Windows command execution.
  * Strengthened workspace path validation for symlinks and junctions.
* **Tests**
  * Expanded lifecycle, integration, concurrency, and acceptance coverage.


whitesarum:
I have an independent data point that supports the bounded-stateful direction of this PR, although it does **not** validate this PR branch or its specific `double markActive` root cause.

After observing multi-GB growth under a high-churn ChatGPT workload, I tested a separate bounded stateful prototype with:

- hard session capacity 256
- explicit active-request and live-SSE protection
- idle-only capacity eviction
- HTTP 503 + `Retry-After: 5` when no eligible idle session exists
- close-failure slots remaining capacity-counted

Real-server validation used Node 24.18.0, MCP SDK 1.30.0 and Express 5.2.1:

- 1,024 / 1,024 real initialize + initialized + DELETE cycles succeeded; final registry/protection occupancy was 0.
- A 256-session capacity run retained one live GET SSE and one disconnected-but-still-running tool operation as protected.
- After the idle-eligibility horizon, 254 eligible sessions were evicted/admitted; the two protected sessions were not evicted.
- The next initialize returned the expected bounded 503 once no eligible idle slot remained.
- Maximum occupancy: 256; active-session evictions: 0; duplicate removal/terminal lifecycle events: 0.
- At equal occupancy 256, post-GC heap after the eviction wave was ~2.08 MB *lower* than the first 256-session checkpoint.
- After full cleanup, post-GC heap was ~1.12 MB above startup.

This suggests that “hard capacity + exact active/stream protection + reject when all slots are protected” is a viable stateful solution shape. The current PR includes many unrelated Windows/tooling changes, so I would still prefer a narrow session-lifecycle PR rather than treating this as validation of the whole branch.

Waishnav:
@gufyhvvyfycyddy-code can you try the latest new stateless architecture which we currently have in v1.1 beta released version. I doubt this will be reproducible now at all :)

https://github.com/Waishnav/devspace/releases#release-v1.1.0-beta.3

PR #91, opened 2026-07-20
Load nested workspace instructions lazily

### LINKED ISSUE #90: open_workspace recursively scans the full workspace and can exceed MCP client timeouts
## Summary

`open_workspace` can block long enough for an MCP client to time out when the requested workspace is a directory containing many independent projects.

The delay comes from eager discovery of nested `AGENTS.md` and `CLAUDE.md` files. This is local filesystem work in DevSpace, not an authentication or tunnel failure.

## Current behavior

`WorkspaceRegistry.createWorkspaceContext()` awaits `findAvailableAgentsFiles(root)` before `open_workspace` responds.

`findAvailableAgentsFiles()` calls `walkWorkspace()`, which recursively and serially traverses every descendant directory except a small fixed skip list. It visits every file entry even though it only needs to identify two instruction-file names. There is no depth, directory-count, or time bound.

## Reproduction

1. Configure an allowed root that contains many project directories and their build/source trees.
2. Call `open_workspace` on that parent directory instead of on one project.
3. Observe that the MCP request can take longer than the client transport timeout.

On one Windows reproduction with roughly 70 direct project folders, the server completed `open_workspace` after about 153 seconds. The MCP client reported a network failure even though the server later logged a successful HTTP 200 response. A single repository opened in about 55 ms; the parent directory still took several seconds after filesystem warming.

## Expected behavior

Opening a workspace should be bounded by the workspace's immediate setup work. Nested instruction discovery should not make the initial MCP request depend on the size of the complete descendant tree.

## Suggested fix: lazy context loading

1. During `open_workspace`, load only root-level and global instruction files.
2. Remove the full descendant walk from the synchronous open path.
3. Before a path-aware tool operates on a file or directory, inspect only that target's ancestor directories up to the workspace root for `AGENTS.md` / `CLAUDE.md`.
4. Return any newly encountered instruction files with that tool response and cache them for the workspace.
5. Apply the same ancestor-context check to shell calls using their explicit working directory.

This preserves nested instruction semantics without silently dropping instructions. A scan timeout alone would avoid the latency but could cause the model to miss applicable instructions.

## Acceptance tests

- Opening a workspace with a large generated/project collection completes within a small fixed budget.
- Root-level instructions are still returned by `open_workspace`.
- A nested instruction file is returned before a tool operates within its directory.
- Previously loaded instruction files are not returned repeatedly.
- Context lookup cannot escape the workspace root through parent traversal or links.

Ahmed-Hindy:
## Summary
- Load nested `AGENTS.md` and `CLAUDE.md` files only when a later path-aware tool enters their directory.
- Cache delivered instruction files and keep realpath containment checks for nested context files.
- Eliminate the recursive workspace walk from `open_workspace`, which could block MCP requests long enough to time out.

## Root cause
`open_workspace` recursively walked every descendant of the selected workspace to build `availableAgentsFiles`. Opening a parent folder that contains many repositories made this synchronous scan dominate the request and caused remote MCP clients to lose the connection.

Closes #90

## Validation
- `npm test`
- `npm run typecheck`
- `npm run build`
- Focused `src/workspaces.test.ts`
- Built runtime benchmark opening a large parent folder: 9 ms, with no eagerly discovered nested instructions

## Summary by CodeRabbit

- **New Features**
  - Automatically provides relevant project instructions when tools enter a directory or operate on files.
  - Applies contextual instructions to reading, writing, editing, shell commands, and patches.
  - Includes instruction details directly in tool results for clearer guidance.

- **Bug Fixes**
  - Prevents repeated instruction files from appearing multiple times.
  - Ignores instruction files accessed through symlinks outside the workspace.


Ahmed-Hindy:
fixes #90

Waishnav:
@Ahmed-Hindy lazily loading nested AGENTS.md files is cool. I thought about the same approach too, but no coding harness currently does that.

They usually let the model know which AGENTS.md files are available and instruct it to read them before it starts working inside those folder paths, so I went with the current approach.

About the issue where a project has a lot of nested AGENTS.md files and discovery can fail because of the MCP timeout: one workaround could be to cap discovery at some threshold and stop searching once we hit it.

you have any other approach for this?

Ahmed-Hindy:
not really. this is what I had in mind and I have been test driving it for a couple of weeks

Waishnav:
ohh, how's the performance of agent loop throughout chatgpt's compaction?

Ahmed-Hindy:
idk. I didn't measure performance that much. Devspace was timing out alot on large repos and was pretty much unusable. I still have alot of issues with it but this specific one is no longer part of it

PR #96, opened 2026-07-21
feat(workflow): add composable workflow execution

Waishnav:
The journal and sandbox from #95 could persist and evaluate scripts, but they did not yet provide the orchestration semantics a workflow author needs: agent calls, concurrency, phases, pipelines, nesting, cancellation, and isolated execution contexts.

Introduce the workflow engine API with `agent()`, `parallel()`, `pipeline()`, `phase()`, `log()`, and nested `workflow()` primitives. Calls share bounded concurrency, propagate phase context, support optional worktree isolation and replay hooks, and run through an injectable provider boundary so engine behavior can be verified without real model calls.

## Stack

Based on #95 and merged into #94 before the CLI and worker layer in #97.

## Summary by CodeRabbit

* **New Features**
  * Added workflow execution with agent calls, parallel processing, pipelines, phases, logging, cancellation, and nested workflows.
  * Added provider selection, concurrency controls, optional worktree isolation, replay support, and structured response extraction.
  * Added workflow error handling and execution metadata.

* **Tests**
  * Added comprehensive integration coverage for concurrency, failures, cancellation, nesting, provider selection, phases, and isolation.
  * Expanded the test command to include the new workflow engine tests.


src/workflow-api.ts
```diff
@@ -0,0 +1,675 @@
...
+      throwIfCancelled(deps);
+
+      if (isolation === "worktree") {
+        if (!deps.createWorktree) {
+          throw new WorkflowEngineError(
+            "worktree",
+            "isolation: 'worktree' requires createWorktree host support",
+          );
+        }
+        worktree = await deps.createWorktree({
```
Waishnav, line 365: [P0] Worktree creation happens before `beginAgentCall`, but the catch path always calls `failAgentCall`. If setup fails (non-Git workspace, missing Git, bad base SHA, collision), the original error is masked by `Unknown workflow agent call`. Begin a `preparing` row before fallible setup, or guard failure journaling with a `callStarted` flag.

src/workflow-api.ts
```diff
@@ -0,0 +1,675 @@
...
+          },
+        });
+        worktree = null;
+      }
+
+      deps.journal.completeAgentCall({
+        runId: deps.runId,
+        callIndex: index,
+        responseText: truncate(result.finalResponse, WORKFLOW_LIMITS.responseTextBytes),
+        structuredJson: structuredJson
```
Waishnav: [P1] Please do not truncate structured JSON. A truncated payload is invalid JSON; replay then silently falls back to response text and changes the return type from object to string. Reject oversized structured results or persist them as an artifact/reference. Only free-form response text should be truncated.

src/workflow-engine.ts
```diff
@@ -0,0 +1,206 @@
...
+    );
+  }
+  const parsed = parseWorkflowScript(input.source, {
+    filename: "workflow:nested",
+  });
+
+  // Child surface: reuse parent agent/parallel/pipeline/phase/log/budget/workflow
+  // so callIndex + semaphore stay shared. Override args + meta for the child body.
+  const childApi: WorkflowApi = {
+    agent: input.parentApi.agent,
```
Waishnav: [P1] Reusing the parent's `agent` closure also reuses the parent's `meta` and phase context, so a child's `defaultProvider` and phase semantics are not actually honored. Share only runtime state (semaphore, call counter, replay, signal, journal), then construct a child API with the child's own `meta` and `args`.

src/workflow-api.ts
Waishnav: Addressed by stacked PR #101 (`d5b00f5`). The agent path now tracks whether `beginAgentCall` actually succeeded and only calls `failAgentCall` for an existing row. A worktree setup failure preserves the original error and still emits a typed `agent_call_failed` event. A regression test makes `createWorktree` throw and verifies that the error is not replaced by `Unknown workflow agent call`. This thread becomes obsolete once #101 is merged.

src/workflow-api.ts
```diff
@@ -0,0 +1,675 @@
...
+  if (typeof record.phase === "string") out.phase = record.phase;
+  if (record.schema !== undefined) {
+    if (!record.schema || typeof record.schema !== "object" || Array.isArray(record.schema)) {
+      throw new WorkflowEngineError("schema", "agent opts.schema must be an object");
+    }
+    out.schema = record.schema as object;
+  }
+  if (typeof record.model === "string") out.model = record.model;
+  if (typeof record.effort === "string") out.effort = record.effort;
+  if (typeof record.provider === "string") out.provider = record.provider;
```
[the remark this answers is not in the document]
Waishnav: Addressed by stacked PR #100 (`4cb4007`). `agent()` options now pass through the strict `agentOptsSchema`, where `provider`, `model`, and `effort` are explicitly typed. Present non-string values fail `safeParse` and are surfaced as `WorkflowEngineError` instead of being silently dropped or falling back to `meta.defaultProvider`. The stacked test/typecheck suite passes on #112.

src/workflow-api.ts
Waishnav: Addressed by follow-up PR #116. Exact agent return values are persisted separately in `return_value_json` and are never text-truncated. `structured_json` remains a bounded inspection field and is omitted rather than stored as malformed partial JSON when oversized. If an exact return value exceeds the replay limit, the live call still succeeds but that call becomes non-replayable and ends the reusable prefix on resume.

src/workflow-engine.ts
Waishnav: Addressed by follow-up PR #117 (`c32b094`). Nested workflows now construct their own `WorkflowApi` from the child script metadata and arguments, while sharing only the runtime semaphore/call index, journal, replay object, cancellation signal, and provider host. A regression test verifies that a parent defaulting to Codex and a child defaulting to Claude invoke the respective providers while retaining one shared call sequence.

src/workflow-api.ts
```diff
@@ -0,0 +1,675 @@
...
+        "nest_depth",
+        `workflow() nesting limited to ${WORKFLOW_MAX_NEST_DEPTH} level`,
+      );
+    }
+    if (!deps.resolveNestedSource || !deps.executeNested) {
+      throw new WorkflowEngineError("internal", "nested workflow() is not configured on this host");
+    }
+    const nameOrRef = args[0] as string | { scriptPath: string };
+    const childArgs = args[1];
+    const source = await deps.resolveNestedSource(nameOrRef);
```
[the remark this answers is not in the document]
Waishnav: Addressed by follow-up PR #117 (`bc387a1`). Explicit nested `{ scriptPath }` references are resolved only inside `<project>/.devspace/workflows`, including canonical realpath containment so symlink escapes are rejected. The loose project `./workflows` named-script fallback was removed; the explicit top-level CLI/MCP script path remains available for cross-harness invocation.

PR #97, opened 2026-07-21
feat(workflow): run and resume workflows from the CLI

Waishnav:
The engine from #96 was only an in-process API. Users still needed a durable way to resolve scripts, launch runs independently of the initiating shell, inspect or cancel them later, isolate agent changes, resume prior work, and request structured output.

Add named and file-based script resolution, per-call workflow worktrees, the initial replay implementation, and JSON Schema enforcement for `agent()` results. A detached CLI worker now owns run execution and heartbeat state, while `workflow run`, `status`, `cancel`, and `ls` provide the first user-facing lifecycle controls.

## Stack

Based on #96 and merged into #94 before the MCP and skill integration in #98.


src/workflow-cli.ts
```diff
@@ -0,0 +1,536 @@
...
+      scriptHash,
+      workspaceRoot,
+      workspaceId: process.env.DEVSPACE_WORKSPACE_ID,
+      argsJson: JSON.stringify(Object.keys(workflowArgs).length ? workflowArgs : null),
+      resumedFromRunId: priorRunId,
+      baseSha,
+    });
+
+    const persisted =
+      priorScriptPath ??
```
Waishnav: [P0] Resume reuses the prior mutable script path and does not verify that the new run is in the same repository/workspace/base SHA. This can replay repository A in B, and `prior.scriptHash` can disagree with edited bytes executed later. Snapshot the exact validated source under the new run ID and require a matching repository identity/base revision (or an explicit override) before replay.

src/workflow-cli.ts
```diff
@@ -0,0 +1,536 @@
...
+async function runWorkflowCancel(args: string[], config: ServerConfig): Promise<void> {
+  const runId = args[0];
+  if (!runId) throw new Error("Usage: devspace workflow cancel <runId>");
+  const store = createWorkflowStore(config);
+  try {
+    const run = store.requestCancel(runId);
+    console.log(formatRunLine(run));
+    if (run.pid && (run.status === "running" || run.status === "starting")) {
+      try {
+        process.kill(run.pid, "SIGTERM");
```
Waishnav: [P0] Sending `SIGTERM` immediately after setting the cooperative flag usually kills the worker before it can observe cancellation, clean up worktrees, or write terminal events. Please centralize CLI/MCP cancellation in a supervisor: request cancel, wait a grace period, then use the cross-platform `terminateProcessTree` fallback and atomically mark the run cancelled. The supervisor should also schedule `reapStale()`.

src/workflow-cli.ts
```diff
@@ -0,0 +1,536 @@
...
+      if (Buffer.byteLength(resultJson, "utf8") > WORKFLOW_LIMITS.resultJsonBytes) {
+        store.failRun(runId, {
+          error: `result exceeds ${WORKFLOW_LIMITS.resultJsonBytes} bytes`,
+          errorKind: "result_too_large",
+        });
+        return;
+      }
+    }
+
+    store.completeRun(runId, { resultJson });
```
Waishnav: [P1] The run becomes `completed` before `run_completed` is appended. A status poll can observe terminal state and stop before the terminal event exists. Please update terminal status and append the corresponding terminal event in one SQLite transaction; use the same path for failed and cancelled runs.

src/workflow-schema.ts
```diff
@@ -0,0 +1,132 @@
...
+  let lastSession: string | undefined;
+  let lastErrors = "unknown validation error";
+
+  for (let attempt = 0; attempt <= maxRetries; attempt += 1) {
+    const prompt =
+      attempt === 0
+        ? basePrompt
+        : `${basePrompt}\n\nPrevious JSON failed validation:\n${lastErrors}\nReturn only corrected JSON.`;
+
+    const result = await input.run(prompt);
```
Waishnav: [P1] `lastSession` is collected but never passed into the next provider call, so every schema retry starts a fresh agent context. Extend the provider input with `providerSessionId` and feed each attempt's returned ID into the next attempt. Also emit `schema_retry` only when another attempt will actually run.

src/workflow-files.test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+{
+  const dir = await mkdtemp(join(tmpdir(), "wf-files-"));
+  const path = await persistWorkflowScript({
+    stateDir: dir,
+    runId: "wfr_test",
+    source: "export const meta = { name: 'x', description: 'd' }\nreturn 1\n",
+    preferredName: "demo",
+  });
+  assert.match(path, /workflow-scripts\/wfr_test\/demo\.js$/);
```
Waishnav: [P1] This assertion is path-separator-specific and is the current Windows CI failure. Please compare normalized path segments, or compare against `join('workflow-scripts', 'wfr_test', 'demo.js')` after normalizing separators.

src/workflow-replay.ts
```diff
@@ -0,0 +1,81 @@
...
+    (row) => row.runId === call.runId && row.callIndex === call.callIndex,
+  );
+  if (idx >= 0) queue.splice(idx, 1);
+  if (queue.length === 0) map.delete(call.cacheKey);
+}
+
+function toHit(call: WorkflowAgentCallRecord): WorkflowReplayHit {
+  if (call.structuredJson) {
+    try {
+      return {
```
Waishnav: [P0] Replay restores only the previous return value; it does not recreate filesystem side effects or reconnect the preserved worktree. A cached mutating agent can report success while the new run contains none of its changes. Please make mutating calls non-cacheable by default, or persist/apply a patch or resume the exact worktree as part of the replay contract.

src/workflow-schema.ts
Waishnav: Addressed by PR #99 (`a8b8e5a`). `providerSessionId` is now part of the workflow provider input, each schema attempt carries forward the latest returned session ID, and the worker passes it into the provider adapter. `schema_retry` is emitted only when another attempt will actually run. Regression coverage verifies that attempt 2 receives the session ID from attempt 1.

src/workflow-files.test.ts
Waishnav: Addressed by PR #99 (`3343404`). The persisted-script assertion now normalizes Windows separators before comparing the path suffix. The Windows smoke job is green on #99, #100, and #101, so this thread is obsolete after the stack lands.

src/workflow-cli.ts
Waishnav: Partially addressed by stacked PR #108 (`c8f0b65`). Resume now reads the selected source, recomputes its hash, persists a fresh per-run script snapshot, updates the new run to that snapshot path, and the worker executes those exact bytes. That fixes the mutable prior-path / stale `scriptHash` half of this finding. The repository identity half is still open: resume does not require matching `workspaceId`/`workspaceRoot`/`baseSha`, so cross-workspace replay is still possible. Keeping this thread unresolved for that remaining check.

src/workflow-cli.ts
Waishnav: Addressed by follow-up PR #115 (`15be40e`, `5a7a3db`). CLI cancellation now uses the shared lifecycle supervisor: set the cooperative flag, wait a grace period, terminate the detached process tree through the cross-platform helper, escalate to SIGKILL when necessary, and atomically mark the run cancelled as a fallback. The server also schedules stale-worker reaping.

src/workflow-cli.ts
Waishnav: Addressed by follow-up PR #115 (`dfbf5a1`). Completed, failed, and cancelled transitions now update the run row and append the matching terminal event inside one immediate SQLite transaction. Pollers can no longer observe terminal status before its terminal event exists.

src/workflow-replay.ts
Waishnav: Addressed by follow-up PR #116 (`57229cb`, `d12fc43`). Resume now reuses only the longest unchanged same-index call prefix. The first mismatch closes replay for every later call. Worktree-isolated calls are explicitly non-replayable until DevSpace can restore their exact worktree, so a cached success can no longer stand in for absent worktree changes. Shared-checkout calls retain Claude Code-like prefix behavior and assume their filesystem effects remain in the checkout.

src/workflow-cli.ts
Waishnav: The mutable-source/hash half was addressed in #108, and PR #116 completes the replay contract with exact return-value persistence plus deterministic same-index prefix reuse. Mandatory workspace/repository identity matching is intentionally not added: `runId` and resume are designed to remain usable from Cursor, Pi, OpenCode, CLI, and other harnesses that do not own a DevSpace workspace ID. Project-local workflow path confinement is handled separately in the next follow-up rather than through workspace-ID authorization.

PR #98, opened 2026-07-21
feat(workflow): expose dynamic workflows through MCP

Waishnav:
Dynamic Workflows were reachable only from the local CLI, so an MCP client could not start or follow durable runs without dropping down to shell commands. The model also lacked focused guidance for choosing workflows and available local-agent providers.

Expose run, status, and cancellation through MCP while keeping execution in the detached worker, so workflows continue after an MCP request disconnects. Add the Dynamic Workflows skill, seed the bundled agent skills consistently, and introduce runtime provider availability to guide workflow execution.

## Stack

Based on #97 and merged into #94 before provider-native structured output in #99.


src/workflow-tools.ts
```diff
@@ -0,0 +1,317 @@
...
+  );
+
+  registerAppTool(
+    server,
+    "workflow_status",
+    {
+      title: "Workflow status",
+      description: "Drain events for a workflow run; optional long-poll yield.",
+      inputSchema: {
+        runId: z.string(),
```
Waishnav, line 229: [P1] Status is authorized only by a global `runId` and is not scoped back to the open workspace. Please include `workspaceId` (or enforce equivalent ownership) and verify the run's `workspaceId`/root before returning events. The same check is needed for cancel and resume.

src/workflow-tools.ts
```diff
@@ -0,0 +1,317 @@
...
+      annotations: { readOnlyHint: false },
+      _meta: {},
+    },
+    async ({ runId }) => {
+      const store = createWorkflowStore(config);
+      try {
+        const run = store.requestCancel(runId);
+        if (run.pid && (run.status === "running" || run.status === "starting")) {
+          try {
+            process.kill(run.pid, "SIGTERM");
```
Waishnav: [P0] MCP cancellation has no grace-period supervision or hard-kill/terminal fallback. It sends one `SIGTERM` and can leave the database row `running` forever if the worker dies before journaling. Please reuse the same shared cancellation supervisor as the CLI and guarantee `run_cancelled` plus terminal status.

Waishnav:
`agentProviders.enabled` is currently applied inside workflow execution, but the rest of the agent surface is inconsistent: `open_workspace` still advertises every live provider and one-shot agent paths can use providers that the user disabled. Please centralize provider policy and reuse it for workflows, `agents run`, `open_workspace`, and `doctor`.

This PR also needs integration coverage for MCP disconnect survival, cancellation reaching a terminal state, paginated terminal event draining, and provider-config filtering.

src/workflow-tools.ts
Waishnav: Addressed by follow-up PR #115. MCP and CLI now call the same `cancelWorkflowRun()` supervisor, including cooperative grace, cross-platform process-tree termination, hard-kill escalation, and an atomic `run_cancelled` terminal fallback. MCP cancellation no longer returns after a single best-effort SIGTERM.

src/workflow-tools.ts
Waishnav: Intentionally not adopting mandatory workspace-scoped status/cancel authorization. DevSpace workflows are also a portable dynamic-workflow layer for Cursor, Pi, OpenCode, and other harnesses that may invoke CLI/MCP without owning a DevSpace `workspaceId`. `runId` remains the durable cross-harness handle by product decision. Filesystem confinement is handled separately through project-local workflow resolution rather than making workspace IDs mandatory.

PR #107, opened 2026-07-25
fix: reopen persisted managed worktrees

### LINKED ISSUE #106: Allow open_workspace to reopen persisted managed worktrees by path
## Problem

A managed worktree created through `open_workspace` is persisted with `mode: "worktree"`, `managed: true`, its worktree root, and an allowed source repository. If the client later loses that workspace ID, reopening the existing worktree by path through `open_workspace` fails with `Path is outside allowed roots` whenever the configured worktree root is intentionally outside `allowedRoots`.

The persisted workspace can still be restored internally by ID, because restore validation separately verifies that its source root is allowed and its worktree root is inside the configured managed-worktree root. There is currently no path-based way to recover that registered workspace.

## Reproduction

1. Configure a source repository under `allowedRoots` and use a separate `worktreeRoot` outside it.
2. Open the source in `worktree` mode and note the returned managed-worktree path.
3. Start a new client session without the returned workspace ID.
4. Call `open_workspace` with the existing managed-worktree path in checkout mode.

### Actual

The call is rejected as outside `allowedRoots`.

### Expected

An exact path match to a persisted managed worktree should restore the original workspace session, including its original workspace ID, without adding the whole worktree parent to `allowedRoots`.

## Security constraints

The recovery must remain limited to a persisted session that is both `mode: "worktree"` and `managed: true`. Normal restored-worktree validation must still require an allowed source root and containment within the configured worktree root. Arbitrary unregistered paths under the managed-worktree parent must remain denied.

## Acceptance criteria

- Reopening a persisted managed worktree by its exact path restores the recorded session.
- Home-directory path expansion is supported by the existing path resolver.
- A regular external checkout and an unregistered sibling worktree remain denied.
- The behavior is covered by a regression test using a managed-worktree root outside `allowedRoots`.

Ahmed-Hindy:
Fixes #106

## Summary

- Let checkout-mode `open_workspace` rehydrate an exact persisted managed worktree by path.
- Reuse the original workspace ID instead of creating a new checkout session.
- Add a regression test with a managed-worktree root outside `allowedRoots`, plus an assertion that an unregistered sibling remains denied.

## Root cause

Persisted managed worktrees already restore safely by workspace ID: their source root is checked against `allowedRoots` and their worktree root against the dedicated managed-worktree root. When the client loses that ID, `open_workspace` only followed the new-checkout path and rejected the same worktree because the worktree parent is intentionally outside `allowedRoots`.

## Security

Rehydration applies only to exact persisted sessions marked as both `mode: "worktree"` and `managed: true`; restored-worktree validation remains unchanged. Arbitrary paths below the managed-worktree parent are still rejected.

## Validation

- `tsx src/workspaces.test.ts`
- `npm run typecheck`
- `npm run build`
- `npm test` reaches an existing Node 24 failure in `src/process-sessions.test.ts:122` before the workspace test runs (`true !== false`).

## Summary by CodeRabbit

* **New Features**
  * Reopen and automatically restore persisted worktree workspaces when they match a managed root.
  * When multiple sessions share the same root, the most recently used session is selected.
  * Restored workspaces keep their identity, mode, source root, and worktree details, including restored agent profiles.

* **Bug Fixes**
  * Prevent opening worktree paths that are not registered or that fall outside approved roots.
  * Improved persistence/reopening behavior for worktree sessions.

### DIFF
--- a/src/workspace-store.ts
+++ b/src/workspace-store.ts
@@ -1,4 +1,4 @@
-import { eq } from "drizzle-orm";
+import { desc, eq } from "drizzle-orm";
 import { openDatabase, type DatabaseHandle } from "./db/client.js";
 import {
   workspaceSessions,
@@ -31,6 +31,7 @@ export interface WorkspaceStore {
     managed?: boolean;
   }): WorkspaceSession;
   getSession(id: string): WorkspaceSession | undefined;
+  getSessionByRoot(root: string): WorkspaceSession | undefined;
   touchSession(id: string): void;
   close?(): void;
 }
@@ -89,6 +90,17 @@ export class SqliteWorkspaceStore implements WorkspaceStore {
       .select()
       .from(workspaceSessions)
       .where(eq(workspaceSessions.id, id))
+      .get();
+
+    return row ? rowToWorkspaceSession(row) : undefined;
+  }
+
+  getSessionByRoot(root: string): WorkspaceSession | undefined {
+    const row = this.database.db
+      .select()
+      .from(workspaceSessions)
+      .where(eq(workspaceSessions.root, root))
+      .orderBy(desc(workspaceSessions.lastUsedAt))
       .get();
 
     return row ? rowToWorkspaceSession(row) : undefined;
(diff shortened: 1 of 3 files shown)


Ahmed-Hindy:
@Waishnav requesting review

Waishnav:
Thanks for the detailed PR and the security-focused regression coverage.

After reviewing this against v1.0.6 and the current workspace contract, we are intentionally not adding path-based rehydration for persisted managed worktrees. v1.0.6 supports conversation-scoped reuse for the same checkout, while every worktree-mode open creates a new managed worktree and workspace. Callers should retain and reuse the `workspaceId` returned by `open_workspace`.

Treating an existing managed-worktree path as a checkout that can be reopened would introduce a second recovery path and blur the checkout/worktree boundary. This is therefore out of scope for the current direction rather than an implementation bug we plan to merge.

Thank you again for the contribution. I am closing this PR as not planned. Issue #106 is the corresponding request and will be closed for the same reason.

PR #116, opened 2026-07-26
fix(workflow): replay deterministic call prefixes

Waishnav:
Resume replay could reuse a later call after an earlier call changed or became unavailable, producing a run assembled from an invalid execution history. Bounded previews were also insufficient as replay values, especially for structured output, and prior worktree filesystem state was not restored automatically.

Persist exact agent return values separately from display previews and replay only the longest unchanged same-index prefix. The first changed, missing, failed, corrupt, or unavailable value closes replay for the remaining calls, while worktree-isolated calls rerun unless their original filesystem state can be restored. Replay remains automatic rather than becoming a public cache option.

## Stack

Based on #115 and merged into #94 before profile and project-script support in #117.

## Summary by CodeRabbit

* **New Features**
  * Workflow resumes now replay only a contiguous, unchanged sequence of prior calls in call order.
  * Exact agent return values are preserved for replay, with a size limit to protect reliability.
  * Replay detects invalid, missing, changed, or unavailable results and continues later calls live.
  * Worktree-based calls are replayed only when the original worktree can be restored.

* **Documentation**
  * Updated dynamic workflow guidance to explain replay behavior and filesystem requirements.


Waishnav:
Added the requested schema-path regression in 8f30225. It verifies an oversized structured result is returned intact to the live workflow while structuredJson is omitted instead of persisting truncated invalid JSON.

PR #147, opened 2026-08-08
fix(server): opt in to root MCP routing for Tailscale Funnel

dbc-hbin:
## Problem

ChatGPT cannot connect to a DevSpace server exposed through a path-based Tailscale Funnel route.

Tailscale Funnel strips the configured path prefix before proxying to the backend. The MCP client requests the documented public `/mcp` endpoint, but a Funnel route targeting that prefix can forward the request to DevSpace as `/`. DevSpace only registers the MCP handler at `/mcp`, so the forwarded request returns 404.

This was verified empirically with a path-echo backend: a request to a configured Funnel prefix arrived with that prefix removed.

## Change

- Add an explicit `DEVSPACE_TAILSCALE_FUNNEL=1` opt-in for path-stripping Funnel routes.
- Pass that integration setting to the server as the provider-neutral `mcpRootAlias` capability.
- Register the authenticated MCP handler at both `/mcp` and `/` only when the capability is enabled.
- Do not infer Funnel mode from a `.ts.net` hostname, because Tailscale Serve uses the same hostname form.
- Document the setting and add an authenticated HTTP integration regression test.

The default remains `/mcp` only. Tailscale Serve and ordinary reverse proxies do not expose the root alias unless explicitly configured.

## Verification

- `npm run typecheck`
- `npx tsx src/config.test.ts`
- `npx tsx src/server.test.ts` — 7/7 pass, including authenticated MCP initialize at `/` and `/mcp`
- `npm run build`

The route regression test verifies:

- opt-in enabled: authenticated MCP initialize at `/` returns 200;
- opt-in disabled: `/` remains 404;
- opt-in disabled: the existing `/mcp` endpoint still returns 200.

## Development attribution

Built with the Codex desktop coding harness using `deepseek/deepseek-v4-flash`. The submitted changes were inspected and validated against the local repository checkout with the commands listed above.

## Summary by CodeRabbit

* **New Features**
  * Added an optional Tailscale Funnel setting that serves MCP requests at the root URL when enabled.
  * MCP requests remain available at `/mcp` regardless of the setting.
  * Added configuration guidance and an example environment variable for enabling the feature.

* **Tests**
  * Added coverage for default behavior, explicit Funnel enablement, Tailscale URLs, regular URLs, and root MCP routing.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -12,6 +12,7 @@ DEVSPACE_ALLOWED_ROOTS=/home/waishnav/personal,/home/waishnav/work
 # For temporary tunnels, prefer setting this per run. DevSpace derives the
 # inbound Host allowlist from this URL.
 # DEVSPACE_PUBLIC_BASE_URL=https://your-public-host.example.com
+# DEVSPACE_TAILSCALE_FUNNEL=1
 # Advanced escape hatch. `*` disables Host header allowlist protection.
 # DEVSPACE_ALLOWED_HOSTS=localhost,127.0.0.1,your-public-host.example.com
 DEVSPACE_TOOL_MODE=full
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -34,10 +34,16 @@ npx @waishnav/devspace config set publicBaseUrl https://devspace.example.com
 | `PORT` | Local port. Defaults to `7676`. |
 | `DEVSPACE_ALLOWED_ROOTS` | Comma-separated local roots that workspaces may open. |
 | `DEVSPACE_PUBLIC_BASE_URL` | Public origin for the server, without `/mcp`. |
+| `DEVSPACE_TAILSCALE_FUNNEL` | Set to `1` when a Tailscale Funnel route strips the `/mcp` path prefix before forwarding. |
 | `DEVSPACE_ALLOWED_HOSTS` | Optional Host header allowlist override. |
 | `DEVSPACE_OAUTH_OWNER_TOKEN` | Owner password for OAuth approval. Must be at least 16 characters. |
 | `DEVSPACE_WORKTREE_ROOT` | Directory for managed Git worktrees. Defaults to `~/.devspace/worktrees`. |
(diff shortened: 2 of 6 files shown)


dbc-hbin:
Addressed the review feedback in 28e7544.

- Removed `.ts.net` hostname inference, so Tailscale Serve remains `/mcp`-only by default.
- Replaced the provider-specific `isTailscaleFunnel` server field with the generic `mcpRootAlias` route capability.
- Added explicit `DEVSPACE_TAILSCALE_FUNNEL=1` opt-in and documentation.
- Added an authenticated HTTP integration test covering `/` enabled, `/` disabled, and unchanged `/mcp` behavior.

Validated with typecheck, focused config/server tests, and a production build.

Waishnav:
Thanks for digging into this — the finding that Tailscale strips a configured path prefix is useful, and the explicit opt-in is better than inferring Funnel from a `.ts.net` hostname.

I don't think we should carry the root MCP alias in DevSpace though. DevSpace exposes more than `/mcp` (`/.well-known/...`, `/authorize`, `/token`, `/register`, `/mcp-app-assets`, etc.), so mounting Funnel specifically at `/mcp` is the underlying configuration problem. The cleaner setup is to proxy the whole DevSpace origin from `/` (for example `tailscale funnel --bg 7676`) and keep the public MCP URL as `https://<host>.ts.net/mcp`.

I'm going to follow up with a small docs change that makes that setup explicit and warns against `--set-path=/mcp`.

Waishnav:
@dbc-hbin Thanks for validating Tailscale’s path-prefix behavior and proposing the root alias. I addressed this in `v1.1.0-beta.1` through documentation in [#285](https://github.com/Waishnav/devspace/pull/285): expose the entire DevSpace origin from the Funnel root with `tailscale funnel --bg 7676`, while keeping the public MCP endpoint at `/mcp`.

Try the beta setup and let me know whether the connection now works end to end.

PR #177, opened 2026-08-10
fix: ChatGPT cannot find allowed workspaces without a root catalog

dbc-hbin:
ChatGPT cannot know an arbitrary local filesystem path before it opens a workspace. With multiple configured roots, it can therefore guess a path outside the DevSpace boundary and fail before it has a usable workspace ID.

Add a read-only `list_allowed_roots` tool that returns only the configured root catalog. Server instructions direct the host to use it when the project path is unknown, and rejected paths now include the same actionable catalog. The change does not scan directories or widen filesystem access.

Validated with `npm run typecheck`, `npx tsx src/roots.test.ts`, `npx tsx src/server.test.ts`, and `npm run build`.

## Development attribution

Built with the Codex desktop coding harness using gpt-5.6-sol. The change was inspected, implemented, and validated against the local repository checkout; the tests listed above were run on the submitted commit.

## Summary by CodeRabbit

* **New Features**
  * Added a tool to list configured workspace names and paths before opening a workspace.
  * Workspace discovery now guides users to identify an available workspace when none is specified.
  * Workspace listings are available in both structured and readable text formats.
* **Bug Fixes**
  * Access-denied messages now include allowed workspace locations, making path restrictions clearer.

### DIFF
--- a/src/roots.test.ts
+++ b/src/roots.test.ts
@@ -1,7 +1,12 @@
 import assert from "node:assert/strict";
 import { homedir } from "node:os";
 import { join, resolve } from "node:path";
-import { assertAllowedPath, expandHomePath, resolveAllowedPath } from "./roots.js";
+import {
+  AccessDeniedError,
+  assertAllowedPath,
+  expandHomePath,
+  resolveAllowedPath,
+} from "./roots.js";
 
 const home = homedir();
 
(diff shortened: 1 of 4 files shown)


Waishnav:
Thanks for the PR, btw can you attach and attribute info about which model/harness you made this PR with?

Waishnav:
tbh, this is not needed to be exposed to the model, but fair enough you have to type out the folder/project path again and again for it's session

you can solve this by creating projects inside ChatGPT and add custom project instructions like this
```
work inside `~/personal/devspace, use DevSpace connector
```
image

dbc-hbin:
Thanks — I added the model/harness attribution to the PR description.

ChatGPT Project instructions are a useful workaround for a stable single-root project. The gap this PR addresses is the generic connector path: a fresh normal chat has no project instructions, and a DevSpace server can expose multiple explicitly configured roots. Before `open_workspace`, the host has no authenticated MCP capability that tells it which local path is valid, so it must guess or ask the user to repeat an implementation detail.

`list_allowed_roots` only returns the roots the owner already configured, to an authenticated client. It does not scan directories, read repository contents, or widen `assertAllowedPath`; `open_workspace` already returns the selected absolute root after opening it. This makes the configured boundary discoverable before the first open rather than expanding that boundary.

I am happy to adjust the name or response shape if you prefer this under a more general read-only server metadata tool.

dbc-hbin:
Addressed the review comment in 8561071. `roots.test.ts` now runs a platform-independent `AccessDeniedError` assertion that checks the complete message contract: the rejected path and every configured allowed root. The existing Windows-specific path case remains covered. Typecheck, roots/server tests, and build pass.

Waishnav:
i'm thinking of adding layer smth similar to `pi` or `opencode`'s way of defining custom tool, so that user can add custom tools like these on their own or this could be added as separate mode or smth, so deferring it for v1.1 release.

PR #193, opened 2026-08-12
perf(acp): reuse Cursor and Copilot connections across sessions

Waishnav:
Cursor and Copilot ACP adapters currently pay process startup and connection setup for each agent turn. This introduces a long-lived ACP runtime per compatible provider identity, performs initialization once, and routes multiple logical ACP sessions over the same connection while keeping their notifications isolated.\n\nCursor and Copilot remain separate runtime identities, and cold continuation still uses the provider session handle when a shared runtime has been evicted. Stacked on #192.

### DIFF
--- a/src/local-agent-acp/config.ts
+++ b/src/local-agent-acp/config.ts
@@ -1,0 +1,114 @@
+export interface AcpSessionConfigState {
+  sessionId: string;
+  configOptions?: unknown;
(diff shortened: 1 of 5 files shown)


Waishnav:
Closing this Sol stack PR in favor of the Luna-based implementation.

PR #201, opened 2026-08-15
feat(mcp): support 2026-07-28 protocol

signal-forge-lab:
DevSpace currently serves the existing sessionful Streamable HTTP flow on /mcp. This adds support for the MCP 2026-07-28 per-request protocol on the same endpoint while preserving existing sessionful clients.

The modern path uses the official @modelcontextprotocol/server and @modelcontextprotocol/node packages. A small adapter reuses the existing DevSpace tool and resource registration instead of duplicating the tool surface, and request metadata continues through to existing handlers.

Verified with npm run typecheck, npm test, npm run build, and git diff --check. I also exercised the branch from a real ChatGPT MCP connection through the public endpoint: tool discovery, open_workspace/read/exec_command, repeated workspace reuse via request metadata, process-session polling/stdin, and tool-error recovery all succeeded. Legacy initialize/session coexistence is covered by the HTTP integration test.

## Summary by CodeRabbit

* **New Features**
  * Added support for the modern MCP protocol alongside existing session-based MCP clients.
  * Added modern MCP discovery, tool and resource access, workspace opening, request metadata, and progress updates over SSE.
  * Added support for OAuth authentication and OpenAI Secure MCP Tunnel connections.
  * Preserved compatibility with legacy MCP requests and sessions.

* **Bug Fixes**
  * Improved request handling, error reporting, and shutdown behavior across MCP connections.

* **Documentation**
  * Updated the README to document supported MCP protocol versions.

PR #280, opened 2026-08-31
fix: bound retained MCP sessions

### LINKED ISSUE #256: 24h MCP session cleanup can be outpaced by high-churn ChatGPT reconnects
## Summary

The 24-hour idle-session cleanup added in #71 is working, but it is not a deterministic memory bound when ChatGPT initializes/reconnects at a high rate. In one long-running production-equivalent run, session creation substantially outpaced 24h cleanup and the DevSpace process grew into multi-GB memory usage.

This is not a claim that MCP session retention is the sole cause of every ChatGPT timeout/disconnect. It is a narrower finding: **the current 24h/5m policy can still allow thousands of retained session/server graphs under high churn.**

## Environment / scope

- macOS
- Node v24.18.0
- ChatGPT as the MCP host
- Observed on a production-equivalent DevSpace v1.0.7-based build with additive read-only runtime observability
- I have **not** installed v1.0.8 to reproduce this exact run. However, current v1.0.8/main still appears to use the same 24h idle timeout, 5-minute cleanup sweep, and no hard session-capacity bound.

## Production evidence

Collector baseline:

- retained MCP sessions: 144
- `heapUsed`: 155,141,692 B
- physical footprint: 229,591,480 B

After about 2h12m:

- retained MCP sessions: 586
- `heapUsed`: 442,791,944 B
- a +256 MiB heap safety threshold was reached

Longer advisory observation before restart:

- current sessions: 3,904
- peak sessions: 4,134
- `heapUsed`: 2,855,774,792 B
- physical footprint: 3,269,620,040 B

The 24h cleanup was actually functioning:

- `idle_timeout` removals: 3,301
- matching `close_completed`: 3,301
- `close_failed`: 0

But sustained creation continued faster than removal. Observed creation rate was roughly 2.5–3.4 sessions/minute. Session count and `heapUsed` were strongly correlated (~0.998 in the focused collector window and ~0.993 over the longer observation). Restart returned the process footprint to roughly 169 MB.

## Interpretation

This supports the original motivation in #71, but suggests the intentionally conservative 24h timeout is insufficient for some ChatGPT workloads. It bounds session *age*, not registry *size* or retained-memory exposure.

A paused conversation vs. memory-safety tradeoff remains: simply shortening the idle timeout can invalidate legitimate paused sessions, while leaving it long permits large accumulation.

## Related upstream work

- #71 — merged stale-session cleanup; 24h timeout / 5m sweep
- #89 — open bounded-stateful approach with capacity/active accounting/503 behavior
- #209 — open stateless Streamable HTTP approach
- #140 — ChatGPT developer-MCP availability symptoms; may be separate and I am not claiming the same root cause

I added the detailed operational evidence to #71, and a separate bounded-stateful validation data point to #89/#209.

## Independent bounded-stateful validation

To test whether a deterministic stateful bound is viable, I separately exercised a prototype with:

- hard capacity: 256 sessions
- active-request and live-SSE protection
- idle-only capacity eviction
- bounded HTTP 503 + `Retry-After: 5` when no eligible idle slot exists
- close-failure slots remaining capacity-counted

Using Node 24.18.0, MCP SDK 1.30.0, and Express 5.2.1 on the real server path:

- 1,024 / 1,024 initialize + initialized + DELETE cycles succeeded
- final registry/permit/protection occupancy: 0
- in a 256-session capacity run, one live GET SSE and one disconnected-but-running tool operation remained protected
- 254 eligible idle sessions were evicted/replaced
- when only protected/non-eligible capacity remained, the next initialize returned the expected 503
- maximum occupancy: 256
- active-session evictions: 0
- duplicate removal/terminal lifecycle events: 0
- equal-occupancy post-GC heap delta after the eviction wave: about -2.08 MB
- after complete cleanup, post-GC heap was about +1.12 MB above startup

This does not prove a bounded-stateful design is preferable to #209 stateless handling; it only shows it is a viable mitigation shape under this test.

## Desired outcome

It would be useful to get maintainer guidance on the intended upstream direction:

1. keep stateful sessions but add a deterministic capacity + active-work protection, or
2. move to request-scoped/stateless Streamable HTTP as in #209.

Either way, I think the current 24h cleanup should be treated as an age-based mitigation rather than a deterministic memory bound for high-churn ChatGPT clients.

Maczuga:
ChatGPT can create replacement MCP sessions without closing the previous transports. The existing 24-hour cleanup limits session age, but not how many sessions stay in memory, so a long-running server can accumulate thousands of transports. This change caps retained sessions at 256 by default and exposes the limit as `server.maxMcpSessions`.

When the limit is reached, DevSpace closes the least recently used idle session before accepting another initialize request. Sessions with active HTTP responses or SSE streams are not evicted. Concurrent initialize requests reserve their slots before transport setup. If every slot is active, initialization returns `503` with `Retry-After: 5` instead of growing the registry.

The tests cover 1,024-session churn, concurrent reservations, synchronous `onclose` removal, close failures, request reference counting, and the authenticated HTTP `503` path with live SSE sessions. `npm test`, `npm run typecheck`, `npm run build`, and `git diff --check` pass.

Fixes #256.

## Summary by CodeRabbit

* **New Features**
  * Added configurable limits for stateful MCP sessions, defaulting to 256.
  * Idle sessions are automatically evicted when capacity is reached, while active sessions remain protected.
  * New session requests receive a `503` response with retry guidance when all retained sessions are active.
  * Improved session lifecycle handling helps prevent access to sessions that are closing.

* **Documentation**
  * Documented the new session capacity setting and runtime behavior.

* **Tests**
  * Added coverage for capacity limits, eviction, concurrent access, and failure scenarios.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -25,6 +25,7 @@ Run `devspace init` to create both files. `devspace config set publicBaseUrl
   "server": {
     "host": "127.0.0.1",
     "port": 7676,
+    "maxMcpSessions": 256,
     // Use the public origin only; do not append /mcp.
     "publicBaseUrl": "https://devspace.example.com",
     "allowedHosts": [],
@@ -76,6 +77,11 @@ Run `devspace init` to create both files. `devspace config set publicBaseUrl
 Omitted sections and keys use the defaults shown above. An empty
 `workspaces.allowedRoots` uses the current working directory. Unknown keys are
 rejected so spelling mistakes cannot silently alter behavior.
+
+`server.maxMcpSessions` limits how many stateful MCP sessions DevSpace keeps in
+memory. When the limit is reached, DevSpace closes the least-recently-used idle
+session first. Active sessions are not evicted. If all retained sessions are
+active, new initialize requests return `503` until a slot is free.
 
 ## Tool modes and UI
 
--- a/schema/v1/devspace.schema.json
+++ b/schema/v1/devspace.schema.json
@@ -29,6 +29,12 @@
           "minimum": 1,
           "maximum": 65535
         },
+        "maxMcpSessions": {
+          "default": 256,
+          "type": "integer",
+          "exclusiveMinimum": 0,
+          "maximum": 9007199254740991
+        },
         "publicBaseUrl": {
           "default": null,
           "anyOf": [
--- a/src/config-schema.ts
+++ b/src/config-schema.ts
@@ -8,6 +8,7 @@ export const DEVSPACE_CONFIG_SCHEMA_URL =
 const serverConfigSchema = z.object({
   host: z.string().trim().min(1).default("127.0.0.1"),
   port: z.number().int().min(1).max(65_535).default(7676),
+  maxMcpSessions: z.number().int().positive().default(256),
   publicBaseUrl: z.string().url().nullable().default(null),
   allowedHosts: z.array(z.string().trim().min(1)).default([]),
   trustProxy: z.boolean().default(false),
--- a/src/config.test.ts
+++ b/src/config.test.ts
@@ -15,6 +15,7 @@ try {
   const defaults = loadConfig(env);
   assert.equal(defaults.host, "127.0.0.1");
   assert.equal(defaults.port, 7676);
+  assert.equal(defaults.maxMcpSessions, 256);
   assert.equal(defaults.publicBaseUrl, "http://127.0.0.1:7676");
   assert.deepEqual(defaults.allowedRoots, [process.cwd()]);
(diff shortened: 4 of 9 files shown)


Waishnav:
Thanks for the reproduction, measurements, and bounded-stateful implementation — they established clearly that age-based cleanup alone does not bound retained MCP state.

With #201 merged, v1.1 now removes retained transport sessions entirely: MCP 2026 per-request handling with stateless compatibility for older clients. That makes the session cap/LRU layer unnecessary in the target architecture.

Closing this as superseded by the stateless architecture.

PR #293, opened 2026-09-02
feat: add MiniMax subagent provider

octo-patch:
Reason: Add MiniMax as a supported local subagent provider with current global and China model configuration.

This adds a native `minimax` target backed by the existing embedded runtime, registers MiniMax-M3 and MiniMax-M2.7 with current context, pricing, modality, and thinking metadata, and selects the authenticated global or China endpoint. It also updates provider discovery, the generated configuration schema, documentation, and focused regression coverage.

Verified with `pnpm typecheck`, `pnpm test`, and `pnpm build`.

## Summary by CodeRabbit

* **New Features**
  * Added MiniMax as a supported local agent provider.
  * Added support for MiniMax-M3 and MiniMax-M2.7 models, including global and China endpoints.
  * Added authentication-based model selection and configurable thinking modes.
  * Added a MiniMax agent example configuration.

* **Documentation**
  * Documented MiniMax provider setup, supported models, endpoints, credentials, and effort settings.

* **Tests**
  * Added coverage for MiniMax availability, registration, authentication, execution, responses, and cleanup.

### DIFF
--- a/docs/agent-profile-schema.md
+++ b/docs/agent-profile-schema.md
@@ -74,6 +74,7 @@ provider: pi
 provider: cursor
 provider: copilot
 provider: grok
+provider: minimax
 ```
 
 Unsupported or custom providers are rejected. DevSpace maps providers to their
@@ -86,6 +87,7 @@ native integration:
 - `cursor`: ACP
 - `copilot`: ACP
 - `grok`: Grok Build ACP (`grok agent stdio`)
+- `minimax`: embedded MiniMax API runtime, with MiniMax-M3 as the default model
 
 Codex is resolved from the user's environment rather than bundled with
 DevSpace. Run `codex login` normally before using it; set `CODEX_COMMAND` when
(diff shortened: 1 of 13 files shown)


Waishnav:
@octo-patch  Thanks for working on this. Since this PR was opened, the subagent architecture on `main` has changed quite a bit, especially around provider config/env handling. (try installing beta release)

I also don’t think MiniMax needs to be a first-class DevSpace provider anymore. Pi can already supports MiniMax natively, and our Pi adapter can select Pi model providers through the profile, e.g. `provider: pi` + `model: minimax/MiniMax-M3`, with credentials passed through Pi’s provider env config.

That keeps DevSpace from duplicating MiniMax endpoints, model metadata, pricing, auth selection, and another wrapper around the Pi runtime.

Could you rework this against latest `main` around the simpler **MiniMax-via-Pi** approach? Ideally this should mostly be docs/example/config coverage rather than a new `MiniMaxLocalAgentDriver`.

Issue #182, opened 2026-08-11
Decouple OAuth issuer and MCP resource URLs for tunnel/proxy deployments


xiepixie:
### Environment & Context

When connecting DevSpace to ChatGPT through OpenAI's official Secure MCP Tunnel, ChatGPT accesses the MCP server through an OpenAI-hosted tunnel resource, while DevSpace's OAuth Authorization Server may remain hosted at a separate public HTTPS origin.

In this setup, the OAuth Authorization Server origin and the externally visible MCP Resource URI are intentionally different.

### Problem

DevSpace currently derives both its OAuth Authorization Server configuration and its MCP resource identity from `publicBaseUrl`.

In `SingleUserOAuthProvider` (`oauth-provider.ts`), authorization and token exchange validate the requested OAuth `resource` using `@modelcontextprotocol/sdk`'s `checkResourceAllowed`.

The `/mcp` route in `server.ts` performs the same resource validation when checking the bearer token.

`checkResourceAllowed` requires the requested and configured resources to have the same origin.

This works when DevSpace itself is exposed directly at `publicBaseUrl`, but fails when the externally visible MCP resource is an OpenAI Secure MCP Tunnel endpoint while the OAuth Authorization Server remains at another origin.

For example:

```text
OAuth Authorization Server:
https://devspace.example.com

MCP Resource:
<OpenAI-hosted Secure MCP Tunnel endpoint>
```

The resulting resource mismatch causes authorization/token exchange failures such as:

```text
Invalid or missing OAuth resource
```

and subsequent MCP requests may fail with:

```text
401 Unauthorized
```

This prevents DevSpace's current OAuth implementation from working cleanly with OpenAI Secure MCP Tunnel without modifying the server.

### Proposed Solution

Decouple the OAuth Authorization Server URL from the accepted MCP Resource URI(s).

For example, DevSpace could support an explicit configuration such as:

```text
DEVSPACE_PUBLIC_BASE_URL=https://devspace.example.com
DEVSPACE_OAUTH_ALLOWED_RESOURCE_URLS=https://<exact-tunnel-resource>
```

or equivalent structured configuration.

The existing `publicBaseUrl` behavior could remain the default for backward compatibility. When additional resource URLs are configured, authorization, token exchange, refresh-token handling, and `/mcp` bearer-token validation would all validate against the same explicit allowlist.

I would prefer explicit resource URIs rather than trusting broad gateway domains, so OAuth audience/resource binding remains narrowly scoped.

[...]

PR #298, opened 2026-09-04
fix(oauth): support exact resource aliases

wcf778:
## Summary

- add `oauth.allowedResourceUrls` for exact alternate MCP resource URLs
- reuse the provider's resource check for authorization and `/mcp` bearer validation
- keep authorization-code and refresh exchanges bound to the resource originally approved
- update the versioned JSON Schema, configuration reference, and regression tests

## Why

OpenAI Secure MCP Tunnel can present a resource URL on the tunnel service rather than the DevSpace `publicBaseUrl` origin. DevSpace currently rejects that mismatch, so the OAuth flow fails closed even though both URLs route to the same operator-controlled server.

This implements the explicit allowlist proposed in #182. Configured aliases use complete URL equality after fragment removal; they do not accept an entire origin or path descendants. The existing canonical `/mcp` behavior remains unchanged.

## Testing

- `pnpm typecheck`
- `node --import tsx --test --test-concurrency=1 src/config-schema.test.ts src/config.test.ts src/oauth-store.test.ts`
- `pnpm build`
- `pnpm test` (96 passed, 3 skipped; the pre-existing Windows failure at `src/process-sessions.test.ts:160` also reproduces on unmodified `main`)

## Summary by CodeRabbit

* **New Features**
  * Added `oauth.allowedResourceUrls` configuration for alternate MCP resource URLs, including secure tunnel and local development addresses.
  * Allowed URLs must use HTTPS, except for localhost and loopback addresses over HTTP.
  * OAuth authorization and token exchanges now require exact resource URL matches, including token refreshes.
  * Removed aliases can no longer mint tokens after a restart.
* **Documentation**
  * Updated configuration guidance with validation rules, automatic public `/mcp` access, and restart requirements after alias changes.


Waishnav:
Thanks for this PR @wcf778  — I prefer this approach over #259. Keeping additional resource aliases as exact URLs and binding token exchange/refresh to the originally authorized resource feels like the right model here.

There are a few things I’d like to tighten before merging:

1. On refresh, can we also verify that the stored/effective resource is still allowed by the current `allowedResourceUrls` policy? Right now a refresh token issued for an alias can still mint a new access token after that alias has been removed from config and DevSpace restarted. `/mcp` will reject it later, but refresh should fail at the policy boundary itself.

2. Please document that changes to `oauth.allowedResourceUrls` require restarting DevSpace, since the provider snapshots this config during server creation.

3. It would be good to restrict configured resource URLs to HTTPS, with HTTP allowed only for explicit loopback addresses. `z.string().url()` currently accepts schemes/remote HTTP URLs that we probably shouldn't accept for bearer-token resources.

Waishnav:
@wcf778 any update?

wcf778:
Thanks for the review @Waishnav — addressed all three points in f7fbc1b.

Refresh now checks the stored resource against the current provider policy before issuing or rotating tokens, even when the request omits `resource`. Missing or removed resources return `invalid_grant`; an explicit resource still has to match the original authorization. The restart regression covers removing an alias, rejecting refresh with omitted/explicit resources, and preserving canonical-resource refresh.

Configured aliases now require HTTPS, with HTTP limited to `localhost`, `127.0.0.1`, or `[::1]` (optional ports). The configuration docs and generated schema describe this restriction, and the docs explicitly require restarting DevSpace after changing `oauth.allowedResourceUrls`.

I also added an HTTP regression through `createServer`: canonical and configured alias tokens initialize MCP successfully, while unconfigured aliases, child paths, and different query strings return 401. Typecheck, build, and all four focused test files pass locally on Windows/Node 24. The full suite reports 97 passed, 3 skipped, and one failure in the unchanged Windows process-interruption test (`src/process-sessions.test.ts:160`, expected the interrupted process to stop); no process/session code was changed in this follow-up.

PR #308, opened 2026-09-06
feat: export workspace files as MCP resources

### LINKED ISSUE #307: Support exporting workspace files as MCP resources
DevSpace currently supports the host-to-workspace direction with `download_artifact`, but there is no corresponding way to hand an existing workspace file back to an MCP host as a native attachment.

I reproduced this with ChatGPT over remote MCP. Returning a `resource_link` whose URI is an HTTPS download URL causes ChatGPT to recognize a file-shaped result, but it does not fetch that URL. Returning a `resource_link` to a registered MCP resource instead (`artifact://...`) causes the host to follow up with `resources/read` and materialize the result as an attachment.

A small implementation could keep this bounded and host-agnostic: add one read-only `export_artifact({ workspaceId, path })` tool, expose files up to a fixed size as short-lived registered MCP resources, return text for textual MIME types and `blob` for binary content from `resources/read`, and reject paths whose real target escapes the selected workspace. This would not require a public artifact HTTP endpoint or a new persistent artifact service.

I have a working local implementation and regression coverage for cross-session resource reads and the size boundary, including a real ChatGPT attachment materialization test. Happy to send the focused PR if this direction fits DevSpace.

svomro:
Closes #307.

DevSpace can already download a native host file into a workspace, but there is no reverse handoff for an existing workspace file. In a real ChatGPT remote-MCP reproduction, an HTTPS `resource_link` was recognized as a file-shaped result but was not fetched; a registered `artifact://` MCP resource was followed with `resources/read` and materialized as a native attachment.

This adds one read-only `export_artifact({ workspaceId, path })` tool. It pins a workspace-contained regular file for five minutes and returns a short-lived `artifact://devspace/<token>` resource link. One `ResourceTemplate` serves follow-up `resources/read` calls, including calls that arrive on a new MCP session. Text resources use MCP text content and binary resources use `blob`. The effective size ceiling is the lower of `artifacts.maxFileBytes` and 8 MiB. The existing Linux-only `download_artifact` behavior is unchanged, and export adds no public artifact HTTP endpoint or persistent artifact store.

The new tests exercise cross-session resource reads, binary blob content, the exact 8 MiB boundary, a lower configured limit, expiry, missing-file error redaction, and symlink escape rejection. The full 110-test suite, typecheck, and production build pass.

## Summary by CodeRabbit

- **New Features**
  - Added support for exporting workspace files to the host through short-lived resource links.
  - Text files are returned as text, while binary files are returned as downloadable content.
  - Added size limits, expiration, and protections against invalid or unsafe file paths.
  - Existing native file download support remains available on supported platforms.

- **Documentation**
  - Updated artifact exchange, configuration, and security documentation to describe file export and download behavior, limits, platform availability, and logging.


Waishnav:
Might explore it in Future, if it adds any value. Thanks anyway

But closing this for now

PR #314, opened 2026-09-06
feat(codex): allow opt-in network access for subagents

xuanmingguo:
Codex subagents currently receive a workspace-write sandbox policy without network access, so tasks that fetch dependencies or call an HTTP endpoint fail even when the owner wants a network-enabled worker. There is no persisted setting to enable networking while retaining the filesystem sandbox.

This adds an optional `networkAccess` boolean to the `codex` entry in `subagents.providers`. It defaults to false, is passed through the daemon and driver into Codex turn sandbox policies, and applies to both new and continued agents. Read-only and full-access policies remain unchanged; other providers reject the setting. The configuration reference and generated JSON schema include the option and daemon restart guidance.

Verified with typecheck, the full test suite (111 passed, 1 skipped), the configuration round-trip test, build, and package-install smoke test. Regression coverage checks omitted/false/true values and new/resumed turns across the existing sandbox modes. An isolated run through the built DevSpace CLI also loaded `networkAccess: true` from JSONC, launched a real Codex subagent, and completed an HTTPS request with HTTP 200; its session recorded workspace-write with network access enabled. No MCP-host UI acceptance was repeated for this change.

## Summary by CodeRabbit

- **New Features**
  - Added an optional `networkAccess` setting for Codex subagent profiles.
  - When enabled, Codex workers using the `workspace-write` sandbox can access the network.
  - The setting defaults to disabled and does not change filesystem write restrictions.
  - Changes take effect after restarting the agent daemon.

- **Validation**
  - Network access settings are validated and limited to Codex providers.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -115,6 +115,7 @@ Subagent providers are explicit. Omitted providers are disabled:
         "enabled": true,
         "model": "gpt-5.4",
         "effort": "high",
+        "networkAccess": false,
       },
       {
         "id": "claude",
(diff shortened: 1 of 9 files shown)


Waishnav:
this shouldn't need to be configuration imo subagent should be allowed to crawl the internet and access to network. Can you just raise the PR only for enabling the codex's subagent network acecss

PR #322, opened 2026-09-07
feat(mcp): add shell-backed file reads

Waishnav:
Adds `tools.fileRead` with a backward-compatible `tool` default and an experimental `shell` mode. Shell mode omits the dedicated `read` tool and routes workspace instruction and skill reads through the active shell tool instead (`exec_cmd` for Codex, `bash` for Claude).\n\nThis also renames the Codex command tool from `exec_command` to `exec_cmd` as a deliberate breaking cleanup with no compatibility alias. This PR is stacked on #320.

## Summary by CodeRabbit

* **New Features**
  * Added configurable file-reading modes: use a dedicated read tool or route file inspection through the active shell.
  * Added `subagents.instructions` support for on-demand or preloaded instructions.
  * Added configurable allowed resource URLs for OAuth connections.
  * Shell-based reading uses `exec_cmd` in Codex mode and `bash` in Claude mode.

* **Documentation**
  * Updated configuration and workflow documentation with new settings and revised field and tool names.

* **Tests**
  * Added coverage for default and shell-based file-reading behavior.

### DIFF
--- a/docs/chatgpt-coding-workflow.md
+++ b/docs/chatgpt-coding-workflow.md
@@ -180,6 +180,11 @@ returns a process session ID when a command is still
 running after its yield window. Use `write_stdin` to poll it, send input, resize
 a PTY, or send Ctrl-C. Set `tty: true` only for commands that need a terminal.
 
+Set `tools.fileRead` to `shell` to omit the dedicated `read` tool. File,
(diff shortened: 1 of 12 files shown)


Waishnav:
WRONG IDEA, cause read tool also let model read images so it doesn't make sense

PR #327, opened 2026-09-08
fix(agents): talk to opencode over the v1 session endpoints

### LINKED ISSUE #302: opencode subagent: SessionRunner throws ModelUnavailableError for a model listed in /config/providers (opencode 1.18.29)
### Environment

- **devspace:** 1.0.8 (npm global, `@waishnav/devspace`)
- **opencode:** 1.18.29 (upgraded from 1.14.30 while debugging)
- **@opencode-ai/sdk** bundled with devspace 1.0.8: 1.18.29
- **OS:** Ubuntu 22.04 (x86_64), Node v24.14.0
- Subagent provider: opencode (enabled in `~/.devspace/config.json`), agent profile with an explicit `model:` in its frontmatter

### What happened

With subagents enabled and a profile that pins a model (e.g. `model: alibaba-token-plan-cn/glm-5.2`), every `devspace agents run <profile> "..."` fails at session drain time with:

```
SessionRunnerModel.ModelUnavailableError: Model unavailable: alibaba-token-plan-cn/glm-5.2
```

The run surfaces as either `PROVIDER_EXECUTION_ERROR: OpenCode agent execution failed.` or `PROVIDER_PROTOCOL_ERROR: OpenCode did not finish the session before the provider timeout.` — the prompt is admitted, but the turn never produces an assistant message.

### What I verified (to rule out the obvious causes)

1. The model **is** present in the spawned server's `/config/providers` — confirmed by querying `http://127.0.0.1:4096/config/providers` on the daemon-spawned instance: `alibaba-token-plan-cn` lists `glm-5.2`, status active.
2. `opencode run -m alibaba-token-plan-cn/glm-5.2 "..."` in the **same project directory** works fine and returns a reply — so auth for the provider is valid.
3. Session creation via SDK v2 succeeds: `client.v2.session.create({ location: { directory: <project> }, agent: "devspace_read_only" | "devspace_allowed", model: { providerID, id } })` returns a session id, and the session row in opencode's sqlite DB stores exactly `{"id":"glm-5.2","providerID":"alibaba-token-plan-cn","variant":"default"}`.
4. The failure happens right after `booting location services` in the opencode log, with `Failed to drain Session ... ModelUnavailableError` — i.e. the server-side SessionRunner rejects the model reference even though that same model id is listed under that provider.
5. Creating the session **without** an explicit model (letting the agent/project default apply) gets past model resolution — the turn then runs and reaches the LLM (in my case it then failed on an unrelated free-tier 429 from the default model). So the problem looks specific to resolving the explicit model ref handed over by devspace, not auth or networking.
6. Tried the model ref with and without `variant` (no variant / `variant: "high"`); both fail the same way.

### Minimal repro

```bash
# ~/.devspace/config.json
{ "subagents": { "enabled": true }, "providers": [ { "id": "opencode", "enabled": true } ] }
```

```markdown
# ~/.devspace/agents/my-profile.md
---
schema: devspace-agent/v1
name: my-profile
description: ...
provider: opencode
model: alibaba-token-plan-cn/glm-5.2
---
```

```bash
cd <some git project under allowedRoots>
devspace agents run my-profile "What language is this project?"
# -> agent run fails; opencode log shows ModelUnavailableError for that model
```

The provider in question is a custom OpenAI-compatible provider defined in `~/.config/opencode/opencode.json` (aliyun token-plan endpoint) — but the same behaviour appeared with `deepseek/deepseek-v4-flash` (official DeepSeek provider), so it does not look provider-specific.

Happy to provide full daemon logs (`~/.local/share/devspace/agentd.log`) or opencode server logs if useful.

yulong-ge:
Fixes #302

## Problem

`opencode serve` (1.18.x) drops provider auth — keys from `~/.local/share/opencode/auth.json` — from LLM requests made on behalf of sessions created through the **v2 API** (`POST /api/session` + `POST /api/session/{id}/prompt`). The provider call 401s, the assistant message lands with `finish: "error"`, and **no error event reaches the session event stream**, so SDK clients watch the turn stall until the provider timeout. This is what broke devspace's opencode subagents (#302). Upstream report: anomalyco/opencode#47888.

The **v1 API** (`/session/*`) on the same serving instance works fine — same providers, same auth:

| Path | Result |
|---|---|
| v2 SDK `session.create` + `session.prompt` | ❌ provider 401, turn stalls |
| v1 SDK `session.create` + `session.prompt` | ✅ completes normally |
| CLI `opencode run -m <provider>/<model>` | ✅ completes normally |

## Change

Switch the opencode runtime from the v2 client to the v1 client, keeping the shared server launcher (`createOpencodeServer`) — only the HTTP surface changes:

- **Create session**: `POST /session?directory=<workspaceRoot>` (directory via query param; agent/model move to the prompt call)
- **Prompt**: `POST /session/{id}/message` with `parts` + `agent` + inline `model` (`{ providerID, modelID }` — note `modelID`, not v2's `id`; an optional `variant` for effort still rides along and is accepted by 1.18.x). Model and agent are supplied per prompt, which also covers resume turns without separate switchModel/switchAgent calls.
- **Messages**: `GET /session/{id}/message` returns the full history (most recent first, no cursor pagination); reversed to chronological order before reuse.
- **Wait**: the v1 prompt blocks until the turn completes, so `waitForOpencodeSession` now only confirms a terminal assistant message (belt-and-braces poll retained with the same timeout).
- **Health**: v1 has no `/health`; the cheap `GET /session` list doubles as the liveness probe.
- **Permissions**: trimmed to the fields the v1 agent schema accepts (`edit`, `bash`, `webfetch`, `external_directory`). The v2-only keys (`read`, `glob`, `grep`, `list`, `task`) have no v1 equivalent; defaults apply.

## Verification

- `tsc -p tsconfig.json --noEmit` clean
- `local-agent-opencode.test.ts` updated to v1-shaped mocks, all green
- **End-to-end against a real `opencode serve` 1.18.29** with an auth.json-backed provider (`zhipuai-coding-plan/glm-5.2`): `devspace agents run opencode-worker "What language is this project?"` → `status: completed` with a correct final response (the model actually read the file), 38s round trip — where the previous v2 path 401'd and timed out.

Once anomalyco/opencode#47888 is fixed upstream, switching back to v2 should be straightforward (this PR is effectively a revert target), but v1 works today.

## Summary by CodeRabbit

* **Compatibility**
  * Updated the local OpenCode integration to support the latest client and server API surfaces.
  * Model and agent selections are now applied directly when sending prompts, including resumed sessions.

* **Reliability**
  * Improved session readiness checks and message retrieval for more consistent local agent operation.

* **Security**
  * Updated available local-agent permissions to align with the current OpenCode permission model.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -242,10 +242,8 @@ jobs:
       - name: Setup npm trusted publishing
         if: steps.publication.outputs.complete != 'true' && steps.publication.outputs.npm_exists != 'true'
         run: |
-          npm install --global npm@11.8.0
-          npm config set registry https://registry.npmjs.org/
           node_version="$(node -p 'process.versions.node')"
-          npm_version="$(npm --version)"
+          npm_version="$(npx --yes npm@11.8.0 --version)"
           echo "Node.js $node_version"
           echo "npm $npm_version"
           node -e 'const [major, minor] = process.versions.node.split(".").map(Number); if (major < 22 || (major === 22 && minor < 14)) process.exit(1)'
@@ -253,13 +251,14 @@ jobs:
             echo "Expected npm 11.8.0 for trusted publishing; resolved npm $npm_version." >&2
             exit 1
           fi
+          npx --yes npm@11.8.0 config set registry https://registry.npmjs.org/
 
       - name: Publish npm package
         if: steps.publication.outputs.complete != 'true' && steps.publication.outputs.npm_exists != 'true'
         env:
           NPM_TAG: ${{ steps.release.outputs.npm_tag }}
           TARBALL: ${{ steps.pack.outputs.tarball }}
-        run: npm publish "$TARBALL" --access public --tag "$NPM_TAG" --provenance
+        run: npx --yes npm@11.8.0 publish "$TARBALL" --access public --tag "$NPM_TAG" --provenance
 
       - name: Publish GitHub release
         if: steps.publication.outputs.complete != 'true'


Waishnav:
@yulong-ge Thanks for digging into this and tracing it upstream.

The v1 fallback makes sense as a temporary mitigation, but I found a few merge-blocking regressions in the current implementation:

* resumed sessions can return a stale previous assistant response;
* effort-only overrides are dropped, and `variant` should be top-level on the v1 prompt;
* `task: "deny"` and the existing permission policy should be preserved — OpenCode 1.18.29’s v1 schema supports those fields despite the older SDK types.

Since `session.prompt()` already blocks and returns the assistant message for that turn, I think we should use that response directly and remove the extra history polling/reordering logic.

With those cleaned up, I’m good with v1 as a temporary bridge until the upstream v2 auth issue is fixed.

PR #336, opened 2026-09-09
feat: Add antigravity acp provider

starlove7:
Adds Google Antigravity as a supported local subagent provider via the Agent Client Protocol (ACP), running alongside existing ACP-backed drivers like Cursor, Copilot, and Grok.
 
Antigravity uses the official `agy_acp_server` binary distributed through the ACP Registry. This integration wires `antigravity` into `AcpLocalAgentDriver` and `AcpRuntime`, adds executable discovery with an  `ANTIGRAVITY_COMMAND` environment override, and registers the provider icon and schema definitions. Model selection (`gemini-3.8-flash`, `gemini-3.1-pro`) maps cleanly to the standard ACP `session/set_config_option` select category.
  
Verified end-to-end against `agy_acp_server` across initialize, new session, prompt streaming, and session continuation (`devspace agents continue`), and all existing unit and config-schema tests pass. 

## Summary by CodeRabbit

* **New Features**
  * Added Google Antigravity as a supported local coding subagent provider.
  * Added provider availability checks, configuration support, executable discovery, and light/dark theme branding for Antigravity.
  * Added support for Grok and Antigravity provider identifiers in configuration schemas.

* **Documentation**
  * Documented running bounded subagent tasks across supported providers.
  * Added Antigravity setup, authentication, executable configuration, and daemon restart guidance.
  * Documented per-provider command and environment configuration.

* **Tests**
  * Added coverage for Antigravity command resolution, availability checks, platform-specific arguments, and workspace execution.


starlove7 (commented):
I reviewed the feedback.

Waishnav:
Thanks for working on this. I tested it with the actual Antigravity ACP binary and the core ACP integration is viable, but I don’t think this branch is worth carrying forward as-is.

A few concrete issues I found:

* ACP drivers are constructed with `options.env` instead of `providerEnv(provider)`, so provider-specific `command`/`env` config is ignored. This also regresses Cursor/Copilot/Grok, not just Antigravity.
* `providerCommandVariable()` has no Antigravity mapping, so `command` config never becomes `ANTIGRAVITY_COMMAND`.
* Availability checks for `agy_acp_server`, while runtime uses `agy_acp_server.par` on Linux/macOS, so availability can report unavailable even when runtime can resolve the binary.
* The macOS test compares `/var/...` and `/private/var/...` literally, which is why CI is failing there.

`main` has since introduced cleaner provider-env and executable-resolution abstractions, so could you rework this from latest `main` and raise a fresh PR instead? Ideally keep it focused on wiring Antigravity into those existing paths and add a small test covering configured command/env through availability and driver creation.

PR #346, opened 2026-09-11
feat: support native artifact downloads on Windows

signal-forge-lab:
download_artifact is currently Linux-only because secure traversal relies on descriptor-anchored filesystem operations. This PR adds Windows support while keeping the current Linux implementation unchanged. On Windows, the workspace root and each destination directory are pinned with native CreateFileW handles via Koffi; reparse points are rejected and delete sharing is omitted so path components cannot be renamed or replaced during transfer. Existing streaming, size/hash verification, same-directory partial files, and hard-link no-overwrite publication remain shared. Windows-specific validation rejects drive-relative or absolute paths, UNC and device namespaces, ADS, reserved device names, invalid characters, and trailing dot or space segments. Koffi is lazy-loaded only on Windows and is included in the pnpm build-script allowlist. This is intentionally narrower than #103 and based on current main: it carries forward the native Windows handle-pinning approach without refactoring the Linux path or adding macOS support. Verified on Windows with pnpm typecheck, the full test suite (131 passed, 6 skipped, 0 failed), and pnpm build. pnpm test:package-install currently fails on Windows with the same devspace.cmd path-not-found error on an untouched current upstream/main checkout, so that failure is pre-existing and not introduced by this branch.

## Summary by CodeRabbit

- **New Features**
  - Native artifact downloads are now supported on Windows.
  - Windows downloads validate paths and protect destination directories against unsafe links and replacement during transfers.
  - Windows file permissions follow the destination directory’s access rules.

- **Bug Fixes**
  - Improved validation rejects unsafe Windows paths, reserved names, invalid characters, and traversal attempts.

- **Documentation**
  - Updated artifact download and security documentation to describe Linux and Windows behavior, while macOS and BSD remain unsupported.

### DIFF
--- a/docs/artifact-exchange.md
+++ b/docs/artifact-exchange.md
@@ -37,8 +37,13 @@ file-object shape, trusted OpenAI download hosts, and redirects before streaming
 Malformed references, unknown fields, absolute paths, traversal, and symlinked
 parents are rejected.
 
-Downloads are streamed under `artifacts.maxFileBytes` and published as
-owner-only files without overwriting an existing destination. The tool is
-currently available on Linux. It is not registered on macOS, Windows, or BSD
-because Node.js does not expose the required descriptor-relative filesystem
-operations there.
+Downloads are streamed under `artifacts.maxFileBytes` and published without
+overwriting an existing destination. The tool is available on Linux and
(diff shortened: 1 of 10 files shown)


Waishnav:
Thanks for the work @signal-forge-lab . I don't have windows running device around, have you manually tested it?

PR #347, opened 2026-09-12
fix: restore PATH/PATHEXT after spreading the provider environment

surelykidding:
Windows stores the search path in the environment block as mixed-case `Path`. `process.env.PATH` resolves case-insensitively, but `localAgentProviderEnvironment` builds each provider's env with a spread, which copies only the original key casing. Every consumer that later reads `env.PATH` from that plain object gets `undefined`: the availability walk reports codex/grok as "executable not found" on the startup banner, and the Codex driver's `resolveCodexCommand` fails before inference with `PROVIDER_UNAVAILABLE: Codex executable was not found.` — while the same probe succeeds from any shell that exports an uppercase `PATH` (e.g. Git Bash), so it only bites standard Windows service environments such as a cmd-launched MCP server.

The fix restores the canonical `PATH` and `PATHEXT` keys after the spread, sourcing values from any casing present in the inherited env and never overwriting an existing key. A regression test covers a `Path`-only inherited env, and I verified end-to-end in a child process whose environment block contains only `Path` that provider driver resolution now finds the codex executable (it returned undefined before).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Windows environment handling by preserving canonical `PATH` and `PATHEXT` variables when inherited with different capitalization.
  - Ensured both original and normalized path entries, along with unrelated environment variables, remain available to local agents.

- **Tests**
  - Added coverage for Windows-style environment variable casing.

### DIFF
--- a/src/local-agent-config.test.ts
+++ b/src/local-agent-config.test.ts
@@ -61,6 +61,19 @@ assert.deepEqual(inherited, {
   OPENAI_API_KEY: "inherited",
   UNCHANGED: "yes",
 });
+{
+  // Windows environment blocks commonly store "Path"; a spread drops the
+  // case-insensitive process.env.PATH lookup that command resolution relies on.
+  const windowsInherited = { Path: "C:\\tools;C:\\Windows", Other: "kept" };
+  const providerEnv = localAgentProviderEnvironment(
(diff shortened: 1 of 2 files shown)


surelykidding:
Misdirected — refiling against the team's repository. The fix itself is unchanged.

PR #353, opened 2026-09-15
fix(mcp): settle the /mcp auth wait when bearer auth writes the response

yulin0629:
## Problem

`app.all("/mcp")` waits for `bearerAuth` like this:

```ts
await new Promise<void>((resolve, reject) => {
  bearerAuth(req, res, (error?: unknown) => { if (error) reject(error); else resolve(); });
});
```

The SDK's `requireBearerAuth` middleware only calls `next()` on success. On 401/403/400 it writes the response itself and returns without calling `next`, so the Promise above never settles. Every rejected request leaves the handler continuation, `req` and `res` reachable for the lifetime of the process. Clients see a normal 401, so nothing looks wrong; the daemon just grows a little on each bad token, which matters once `/mcp` is on a public hostname.

## Fix

Also resolve when the response finishes or closes, so a request that the middleware rejects releases its handler. Behaviour on success is unchanged (`res.headersSent` short-circuit still applies).

## Verification

- `tsc --noEmit` clean
- `src/server-oauth.test.ts` and `src/server.test.ts` pass (23/23)

Found while running a fork of DevSpace with a static bearer path in front of the same middleware; the bug itself is independent of that change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved request handling so authentication processing completes when a response finishes or closes, preventing requests from remaining pending.


yulin0629:
Closes #354.

Waishnav (changes requested):
Thanks for catching the unresolved auth wait — that part is real. I don’t think we should fix it by resolving on `finish`/`close`, though.

`requireBearerAuth` is already Express middleware, so the simpler and safer fix is to compose it directly:

```ts
app.all("/mcp", bearerAuth, async (req, res) => {
  // existing MCP route logic
});
```

That removes the manual Promise entirely. It also avoids the `close` race where the route can resume before token verification has completed.

I also wasn’t able to reproduce the claimed permanent per-request memory leak: the pending continuation becomes unreachable and the request objects were collectible after GC. So I’d describe the bug as incorrect/unnecessary async control flow rather than a confirmed process-lifetime leak.

Could you rework this PR around native Express middleware composition instead?

PR #362, opened 2026-09-18
fix(codex): fall back to stream items when turn.items is empty

### LINKED ISSUE #361: codex subagent fails with "did not return a final assistant response" when app-server returns empty turn.items
## Summary

The `codex` subagent fails on every run with:

```
PROVIDER_PROTOCOL_ERROR: Codex did not return a final assistant response.
```

`codex exec` works fine on the same machine, so the task and the Codex install are not the problem. The final assistant text is dropped while parsing the app-server turn result.

## Environment

- devspace 1.0.8
- codex-cli 0.139.0 (`codex app-server`)
- Node v22.22.2, Linux (WSL2)

## Measured

Driving `codex app-server` directly over stdio, the `turn/completed` notification carries an empty `items` array:

```
EVENT: item/completed item.type=agentMessage text="CODEXOK"
EVENT: turn/completed
=== turn: {"id":"01a0b00b-a7e7-7de0-85be-f117f07191bf","items":[],"itemsView":"notLoaded","status":"completed","error":null,"startedAt":1789659949,"completedAt":1789659958,"durationMs":8764}
=== item types:
```

So the assistant text only arrives as `item/completed` events (`item.type = "agentMessage"`). The `turn.completed` payload itself has `items: []` and marks `itemsView: "notLoaded"`.

## Cause in the current code

`parseCompletedTurn` in `src/local-agent-codex.ts`:

```ts
const completedItems = (Array.isArray(turn?.items) ? turn.items : items).slice(-MAX_TURN_ITEMS);
```

`turn.items` is an empty **array**, which is truthy, so it wins over the items collected from the event stream. `finalResponse` stays `""`, and the run then fails the `if (!parsed.finalResponse.trim())` check.

## Runtime result

```
$ devspace agents run codex "reply with exactly: CODEXOK" --json
{"id":"agt_ac56c573","status":"running"}

$ devspace agents show agt_ac56c573 --json     # after ~9s
{"id":"agt_ac56c573","status":"failed","error":{"code":"PROVIDER_PROTOCOL_ERROR","message":"Codex did not return a final assistant response.","retryable":false}}
```

And in `devspace agents daemon logs`:

```json
{"event":"agent_run_failed","provider":"codex","agentId":"agt_ac56c573","durationMs":8951,"errorCode":"PROVIDER_PROTOCOL_ERROR","error":"Codex did not return a final assistant response.","causeType":"object","persistenceFailed":false}
```

## Repro

A minimal app-server probe is enough:

1. spawn `codex app-server`
2. `initialize`
3. `thread/start` (cwd: any trusted directory, `approvalPolicy: "never"`, `sandbox: "workspace-write"`)
4. `turn/start` with `input: [{ type: "text", text: "reply with exactly: OK" }]`
5. observe `item/completed` carrying the agent text, then `turn/completed` with `turn.items === []` and `turn.itemsView === "notLoaded"`

## Notes

- The same line exists on current `main`, so this is not specific to the published 1.0.8 build.
- `devspace agents run opencode ...` completes normally on the same machine, so the driver plumbing and turn-completion handling are otherwise working.
- I did not find an existing issue covering this. If there is one, feel free to close this.

Leaving the fix approach to you.

rome-xi:
## Summary

Codex app-server can finish a turn with `turn.items: []` and `itemsView: "notLoaded"` while the assistant text only arrived on `item/completed` (`agentMessage`). `parseCompletedTurn` treated any array `turn.items` as authoritative, so an empty array discarded the stream-collected items and the run failed with `PROVIDER_PROTOCOL_ERROR: Codex did not return a final assistant response.`

This prefers non-empty `turn.items` and otherwise uses the items collected from the event stream. I hit this on 1.0.8 with codex-cli 0.139.0 while driving `devspace agents run codex` for a short smoke prompt; `codex exec` on the same machine was fine, so the drop was in the app-server turn parse path. `v1.1.0-beta.4` still has the same empty-array preference.

Fixes #361

## Test plan

- [x] `pnpm exec tsx --test --test-concurrency=1 src/local-agent-codex.test.ts`
- [x] New regression: mock app-server emits `item/completed` then `turn/completed` with `items: []` / `itemsView: "notLoaded"` → `finalResponse === "CODEXOK"`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of completed turns when item details are unavailable in the final response.
  - Streamed agent responses are now correctly preserved and returned during runtime execution.
  - Ensured only the most recent turn items are processed when multiple items are available.


Waishnav:
Thanks for the clear repro and focused fix — confirmed locally against Codex 0.139.0 and newer. Much appreciated!

